PR #13, opened 2016-02-18
README code example

gsquire:
The README had some minor fixes that were needed in order for it to compile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,14 +25,14 @@ mod hello_service {
 struct HelloService;
 impl hello_service::Service for HelloService {
     fn hello(&self, name: String) -> String {
-        format!("Hello, {}!", s)
+        format!("Hello, {}!", name)
     }
 }
 
 fn main() {
     let server_handle = hello_service::serve("0.0.0.0:0", HelloService, None).unwrap();
     let client = hello_service::Client::new(server_handle.local_addr(), None).unwrap();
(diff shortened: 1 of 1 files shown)


tikue:
Thanks so much for catching that! I think we can even remove the first `to_owned` entirely, because `&str`s can be compared to `String`s.

killercup:
I recently saw https://github.com/brson/rust-skeptic which runs test in markdown files likes Readme.mds. Maybe you could use that to prevent cases like these :)

tikue:
@killercup that's amazing, thank you. brson is a hero as usual.

README.md
```diff
@@ -25,14 +25,14 @@ mod hello_service {
...
-        format!("Hello, {}!", s)
+        format!("Hello, {}!", name)
     }
 }
 
 fn main() {
     let server_handle = hello_service::serve("0.0.0.0:0", HelloService, None).unwrap();
     let client = hello_service::Client::new(server_handle.local_addr(), None).unwrap();
-    assert_eq!("Hello, Mom!".into(), client.hello("Mom".into()).unwrap());
+    assert_eq!("Hello, Mom!".to_owned(), client.hello("Mom".to_owned()).unwrap());
```
tikue: Change to `assert_eq!("Hello, Mom!", client.hello("Mom".into()).unwrap());`

tikue:
Accepting this as-is because I don't want to have a broken readme. Thanks again!

PR #14, opened 2016-02-18
Hooks

shaladdle:
## `pre-push`

1. Check that the working copy is clean
2. Check for `multirust`
3. If `multirust` is present, run tests on various toolchains
4. If it's not present, just run tests with current toolchain

## `pre-commit`

1. Check for non-ascii filenames
2. Check for bad whitespace in files
3. Check for existence of rustfmt + shasum
4. If they are present, check that files being committed are properly formatted

### DIFF
--- a/hooks/pre-push
+++ b/hooks/pre-push
@@ -40,55 +40,67 @@ SUCCESS="${GREEN}ok${NC}"
 
 printf "${PREFIX} Clean working copy ... "
 git diff --exit-code &>/dev/null
-if [ "$?" == 0 ];
-then
+if [ "$?" == 0 ]; then
 	printf "${SUCCESS}\n"
 else
 	printf "${FAILURE}\n"
 	if [ "${TARPC_ALLOW_DIRTY}" == "1" ]
 	then
-		echo ${WARNING} Running with unclean working directory
+		printf "${WARNING} Running with unclean working directory\n"
 	else
 		exit 1
 	fi
 fi
 
-TEST_RESULT=0
+PREPUSH_RESULT=0
 
 # args:
 # 1 - cargo command to run (build/test)
 # 2 - directory name of crate to build
 # 3 - rust toolchain (nightly/stable/beta)
 function run_cargo() {
-	if [ "$1" == "build" ]
-	then
+	if [ "$1" == "build" ]; then
 		VERB=Building
 	else
 		VERB=Testing
 	fi
-	if [ "$3" != "" ]
-	then
+	if [ "$3" != "" ]; then
 		printf "${PREFIX} $VERB $2 on $3 ... "
 		multirust run $3 cargo $1 --manifest-path $2/Cargo.toml &>/dev/null
 	else
 		printf "${PREFIX} $VERB $2 ... "
 		cargo $1 --manifest-path $2/Cargo.toml &>/dev/null
 	fi
-	if [ "$?" != "0" ];
-	then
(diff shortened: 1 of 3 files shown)


shaladdle:
I signed it!

hooks/pre-push
```diff
@@ -10,25 +33,80 @@ GREEN='\033[0;32m'
...
+printf "${PREFIX} Clean working copy ... "
 git diff --exit-code &>/dev/null
-if [ "$?" != 0 ];
+if [ "$?" == 0 ];
 then
-	echo ${RED}ERROR${NC} You have uncommitted changes please commit or stash them before pushing so that I can run tests!
-	exit 1
+	printf "${SUCCESS}\n"
+else
+	printf "${FAILURE}\n"
```
tikue: Looks like this printf should go in the else block

tikue:
So, build on all 3 trains, but only test on nightly?

hooks/pre-push
```diff
@@ -10,25 +33,80 @@ GREEN='\033[0;32m'
...
+
+printf "${PREFIX} Checking for multirust ... "
+command -v multirust &>/dev/null
+if [ "$?" == 0 ] && [ "${TARPC_USE_CURRENT_TOOLCHAIN}" == "" ];
 then
-	printf "${RED}FAILED${NC}"
-	TEST_RESULT=1
+	printf "${SUCCESS}\n"
+
+	# No need to build on nightly, the tests will do that
```
tikue: We do need to build on nightly, because there are dev dependencies only built when testing, so there are failure modes only when building, e.g. if an `extern crate` isn't `#[cfg(test)]`ed but it's listed under `[dev-dependencies]`

shaladdle:
- Now building on nightly too
- Fixed that printf
- Fixed some minor other things

[sample pre-push output](https://www.dropbox.com/s/k20z2gwigzuydis/Screen%20Shot%202016-02-18%20at%201.45.20%20AM.png?dl=0), [sample pre-commit output](https://www.dropbox.com/s/4bsty93u2m7z13t/Screen%20Shot%202016-02-18%20at%201.49.23%20AM.png?dl=0)
[me writing bash](http://img1.rnkr-static.com/user_node_img/50013/1000240969/870/doing-science-photo-u1.jpg)

hooks/pre-push
```diff
@@ -10,30 +33,95 @@ GREEN='\033[0;32m'
...
+	run_cargo build tarpc_examples nightly
+
+	# We still rely on some nightly stuff for tests
+	run_cargo test tarpc nightly
+	run_cargo test tarpc_examples nightly
 else
-	printf "${GREEN}ok${NC}"
-fi
-printf "\n"
+	printf "${FAILURE}\n"
```
tikue: I don't know if success/failure should be printed when checking for the toolchain, since we proceed in either case. Isn't warning enough?

tikue:
Works great for me! One question -- I'm a bit concerned about having to remember to `cp hooks/* .git/hooks` whenever there are changes. Could we instead have the thing in `.git/hooks` delegate to the `hooks/` directory?

hooks/pre-push
tikue: "ok" is good for success, but maybe just not "FAILURE" when a toolchain is missing

shaladdle:
Changed to print NOT FOUND instead of FAILURE in the case that multirust isn't found. If you want the scripts to auto update, just symilnk instead of copying.

PR #17, opened 2016-02-18
Define RPC in readme

tikue:


tikue:
Rendered: https://github.com/tikue/tarpc/tree/readme

PR #20, opened 2016-02-19
[Breaking] Renamings and refactorings.

### LINKED ISSUE #18: Consider new constructor functions for Client
`Client::new("127.0.0.1:9000", None)` seems pretty cryptic if you don't remember what the type signature is. I think it would make sense to provide the following functions instead:

```
fn new<A: ToSocketAddrs>(addr: A) -> ...
fn with_timeout<A: ToSocketAddrs>(addr: A, timeout: Duration) -> ...
```

In my opinion, this makes the intent of the programmer more clear, especially in cases when you don't want a timeout.

If you like the idea I could submit a PR. This would probably require bumping the version of the library to 0.3.0 though.

tikue:
* Add a Config struct. `Client::new(addr, duration)` => `Client::new(addr, config)`
* `serve(addr, server, config)` => `server.spawn_with_config(addr, config)`
* Also added `server.spawn(addr)`

Fixes #18

### DIFF
--- a/tarpc/src/macros.rs
+++ b/tarpc/src/macros.rs
@@ -218,17 +218,22 @@ macro_rules! impl_deserialize {
 /// # }
 /// ```
 ///
+/// There are two rpc names reserved for the default fns `spawn` and `spawn_with_config`.
+///
 /// Attributes can be attached to each rpc. These attributes
 /// will then be attached to the generated `Service` trait's
 /// corresponding method, as well as to the `Client` stub's rpcs methods.
 ///
 /// The following items are expanded in the enclosing module:
 ///
-/// * `Service` -- the trait defining the RPC service
+/// * `Service` -- the trait defining the RPC service. It comes with two default methods for
+///                starting the server:
+///                1. `spawn` starts the service in another thread using default configuration.
+///                2. `spawn_with_config` starts the service in another thread using the specified
+///                   `Config`.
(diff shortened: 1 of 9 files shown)


tarpc/src/macros.rs
```diff
@@ -38,7 +38,10 @@ macro_rules! client_methods {
             if let __Reply::$fn_name(reply) = reply {
                 ::std::result::Result::Ok(reply)
             } else {
-                panic!("Incorrect reply variant returned from protocol::Clientrpc; expected `{}`, but got {:?}", stringify!($fn_name), reply);
+                panic!("Incorrect reply variant returned from protocol::Clientrpc; expected `{}`, \
```
shaladdle: s/Clientrpc/Client.rpc?

tarpc/src/macros.rs
```diff
@@ -76,7 +79,10 @@ macro_rules! async_client_methods {
                 if let __Reply::$fn_name(reply) = reply {
                     reply
                 } else {
-                    panic!("Incorrect reply variant returned from protocol::Clientrpc; expected `{}`, but got {:?}", stringify!($fn_name), reply);
+                    panic!("Incorrect reply variant returned from protocol::Clientrpc; expected \
```
shaladdle: Same as above

tikue:
It'll be OK, coverage.

PR #22, opened 2016-02-20
Allow most things to be unused

shaladdle:
I noticed that I there was a painfully high number of warnings from the generated items if you don't use them right away. Even if you do use some of them, you might want to use just `Client` or `AsyncClient`, and currently you'll get warnings for the one you don't use.


tikue:
I don't love this solution, but I think the current alternative is worse. Perhaps we should consider expanding the macro at some point to allow opting out of certain things, like `AsyncClient` and `Future`, or even just creating a `Server` XOR `Client`.

tikue:
(Of course, that doesn't solve the awful first-use experience you'll face from the unused warnings)

PR #27, opened 2016-02-23
[Breaking] Add support for arbitrary transports.

### LINKED ISSUE #16: Make it transport-agnostic
I see you've said on reddit that you're thinking about making this library transport-agnostic. Here's my two cents:

[urpc](https://github.com/kmcallister/urpc) can listen on a single `Read + Write` object, which is perfect for my use case – I have a library called [rusty-sandbox](https://github.com/myfreeweb/rusty-sandbox) which creates a pipe between two processes and gives access to it as a `File`. Please add support for this!

tikue:
Quite a bit of machinery added:
 * Listener trait
 * Dialer trait
 * Stream trait
 * Transport trait

I made a lot of types default to `Tcp*` for:
 1. Convenience
 2. Backwards compatibility (though it's still a breaking change in some respects)

~~Before merging this, we should `impl` the traits for [unix_socket](http://rust-lang-nursery.github.io/unix-socket/doc/v0.5.0/unix_socket/) to make sure the trait is compatible.~~
I've now `impl`ed `Transport` for  Unix sockets:

* `Server.spawn(UnixTransport(path))`
* `Client::new(UnixDialer(path))`

**This is a breaking change** but only for code that used `ToSocketAddrs` args other than `&str`. Code using `&str` to specify address will still work.
To update code:
`Server.spawn(socket_addr)` => `Server.spawn(TcpTransport(socket_addr))`

### Bikeshed
As I played around with unix sockets, I grew frustrated with the second-class nature of them WRT tarpc. Specifically, having to use `spawn_with_config` just because the transport changed, even though I was using default `Config` settings, was annoying. In light of that, I propose making `spawn` take a `Transport`, and just go ahead and `impl<A: ToSocketAddrs> Transport for A { ... }` such that it creates a tcp transport. Even though other transports use sockets, I think it's reasonable to have a sane default here for the sake of ergonomics.

If you think that's reasonable, I'll add that to this PR, as well.

*Update: the above was implemented as well, but due to coherence, I can only do `impl<'a> Transport for &'a str`. Specialization might allow `Transport` impls for all `ToSocketAddrs` types, but maybe not.*

Fixes #16

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,17 +56,18 @@ fn main() {
 
 The `service!` macro expands to a collection of items that collectively form an rpc service. In the
 above example, the macro is called within the `hello_service` module. This module will contain a
-`Client` (and `AsyncClient`) type, and a `Service` trait. The trait provides `default fn`s for
-starting the service: `spawn` and `spawn_with_config`, which start the service listening on a tcp
-port. A `Client` (or `AsyncClient`) can connect to such a service. These generated types make it
-easy and ergonomic to write servers without dealing with sockets or serialization directly. See the
-tarpc_examples package for more sophisticated examples.
+`Client` (and `AsyncClient`) type, and a `Service` trait. The trait provides default `fn`s for
+starting the service: `spawn` and `spawn_with_config`, which start the service listening over an
+arbitrary transport. A `Client` (or `AsyncClient`) can connect to such a service. These generated
+types make it easy and ergonomic to write servers without dealing with sockets or serialization
+directly. See the tarpc_examples package for more sophisticated examples.
 
 ## Documentation
 Use `cargo doc` as you normally would to see the documentation created for all
 items expanded by a `service!` invocation.
 
 ## Additional Features
+- Connect over any transport that `impl`s the `Transport` trait.
 - Concurrent requests from a single client.
 - Any type that `impl`s `serde`'s `Serialize` and `Deserialize` can be used in the rpc signatures.
 - Attributes can be specified on rpc methods. These will be included on both the `Service` trait
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ impl HelloService for HelloServer {
 
 fn main() {
     let server_handle = HelloServer.spawn("0.0.0.0:0").unwrap();
-    let client = hello_service::Client::new(server_handle.local_addr()).unwrap();
+    let client = hello_service::Client::new(server_handle.dialer()).unwrap();
     assert_eq!("Hello, Mom!", client.hello("Mom".into()).unwrap());
     drop(client);
     server_handle.shutdown();
--- a/README.md
+++ b/README.md
@@ -46,8 +46,9 @@ impl HelloService for HelloServer {
 }
 
 fn main() {
-    let server_handle = HelloServer.spawn("0.0.0.0:0").unwrap();
-    let client = hello_service::Client::new(server_handle.dialer()).unwrap();
+    let addr = "localhost:10000";
+    let server_handle = HelloServer.spawn(addr).unwrap();
+    let client = hello_service::Client::new(addr).unwrap();
     assert_eq!("Hello, Mom!", client.hello("Mom".into()).unwrap());
     drop(client);
     server_handle.shutdown();
--- a/tarpc/src/macros.rs
+++ b/tarpc/src/macros.rs
@@ -522,6 +522,8 @@ mod syntax_test {
 #[cfg(test)]
 mod functional_test {
     extern crate env_logger;
+    use Config;
+    use transport::unix::UnixTransport;
 
     service! {
         rpc add(x: i32, y: i32) -> i32;
@@ -577,6 +579,20 @@ mod functional_test {
         let client2 = client1.try_clone().unwrap();
         assert_eq!(3, client1.add(1, 2).get().unwrap());
         assert_eq!(3, client2.add(1, 2).get().unwrap());
+    }
+
+    #[test]
+    fn async_try_clone_unix() {
+        let handle = Server.spawn_with_config(UnixTransport("/tmp/test"),
+                                              Config::default()).unwrap();
+        let client1 = AsyncClient::with_config(handle.dialer(),
+                                               Config::default()).unwrap();
+        let client2 = client1.try_clone().unwrap();
+        assert_eq!(3, client1.add(1, 2).get().unwrap());
+        assert_eq!(3, client2.add(1, 2).get().unwrap());
+        drop(client1);
+        drop(client2);
+        handle.shutdown();
     }
 
     // Tests that a server can be wrapped in an Arc; no need to run, just compile
--- a/tarpc/src/macros.rs
+++ b/tarpc/src/macros.rs
@@ -386,7 +386,8 @@ macro_rules! service_inner {
 
         #[allow(unused)]
         #[doc="The client stub that makes RPC calls to the server."]
-        pub struct Client<S: $crate::Stream>($crate::protocol::Client<__Request, __Reply, S>);
+        pub struct Client<S = ::std::net::TcpStream>($crate::protocol::Client<__Request, __Reply, S>)
+            where S: $crate::Stream;
 
         impl Client<::std::net::TcpStream> {
             pub fn new<A>(addr: A) -> $crate::Result<Self>
@@ -396,7 +397,9 @@ macro_rules! service_inner {
             }
         }
 
-        impl<S: $crate::Stream> Client<S> {
+        impl<S> Client<S>
+            where S: $crate::Stream
+        {
             #[allow(unused)]
             #[doc="Create a new client with default configuration that connects to the given \
                    address."]
@@ -427,7 +430,8 @@ macro_rules! service_inner {
 
         #[allow(unused)]
         #[doc="The client stub that makes asynchronous RPC calls to the server."]
-        pub struct AsyncClient<S: $crate::Stream>($crate::protocol::Client<__Request, __Reply, S>);
+        pub struct AsyncClient<S = ::std::net::TcpStream>($crate::protocol::Client<__Request, __Reply, S>)
+            where S: $crate::Stream;
 
         impl AsyncClient<::std::net::TcpStream> {
             #[allow(unused)]
@@ -440,7 +444,8 @@ macro_rules! service_inner {
             }
         }
 
-        impl<S: $crate::Stream> AsyncClient<S> {
+        impl<S> AsyncClient<S>
+            where S: $crate::Stream {
             #[allow(unused)]
             #[doc="Create a new asynchronous client that connects to the given address."]
             pub fn with_config<D>(dialer: D, config: $crate::Config) -> $crate::Result<Self>
@@ -466,7 +471,8 @@ macro_rules! service_inner {
         }
 
         #[allow(unused)]
-        struct __Server<S: 'static + Service>(S);
+        struct __Server<S>(S)
+            where S: 'static + Service;
 
         impl<S> $crate::protocol::Serve for __Server<S>
             where S: 'static + Service
--- a/tarpc/src/macros.rs
+++ b/tarpc/src/macros.rs
@@ -316,17 +316,17 @@ macro_rules! service_inner {
 
             #[doc="Spawn a running service."]
             fn spawn<A>(self, addr: A)
-                -> $crate::Result<$crate::protocol::ServeHandle<$crate::TcpDialer<::std::net::SocketAddr>>>
+                -> $crate::Result<$crate::protocol::ServeHandle<$crate::transport::tcp::TcpDialer<::std::net::SocketAddr>>>
                 where A: ::std::net::ToSocketAddrs,
                       Self: 'static,
             {
-                self.spawn_with_config($crate::TcpTransport(addr), $crate::Config::default())
+                self.spawn_with_config($crate::transport::tcp::TcpTransport(addr), $crate::Config::default())
             }
 
             #[doc="Spawn a running service."]
             fn spawn_with_config<T>(self, addr: T, config: $crate::Config)
-                -> $crate::Result<$crate::protocol::ServeHandle<<T::Listener as $crate::Listener>::Dialer>>
-                where T: $crate::Transport,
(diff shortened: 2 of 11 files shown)


tarpc/src/macros.rs
```diff
@@ -569,6 +581,20 @@ mod functional_test {
         assert_eq!(3, client2.add(1, 2).get().unwrap());
     }
 
+    #[test]
+    fn async_try_clone_unix() {
+        let handle = Server.spawn_with_config(UnixTransport("/tmp/test"),
```
tikue: This won't work if you run it twice because the file already exists. We can just delete it at the end I guess?/Choosing something like /tmp/tarpc-async_try_clone_UNIX?

tarpc/src/transport/unix.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+impl super::Listener for UnixListener {
+    type Stream = UnixStream;
+    type Dialer = UnixDialer<PathBuf>;
+    fn accept(&self) -> io::Result<UnixStream> {
+        self.accept().map(|(stream, _)| stream)
+    }
+    fn dialer(&self) -> io::Result<UnixDialer<PathBuf>> {
+        self.local_addr().map(|addr| UnixDialer(addr.as_pathname().unwrap().to_owned()))
```
tikue: I shouldn't be unwrapping here, and we shouldn't merge as-is. I just haven't thought about the right thing to do yet. Thoughts?

tarpc/src/macros.rs
```diff
@@ -569,6 +581,20 @@ mod functional_test {
...
+                                              Config::default()).unwrap();
+        let client1 = AsyncClient::with_config(handle.dialer(),
+                                               Config::default()).unwrap();
+        let client2 = client1.try_clone().unwrap();
+        assert_eq!(3, client1.add(1, 2).get().unwrap());
+        assert_eq!(3, client2.add(1, 2).get().unwrap());
+        drop(client1);
```
shaladdle: I guess I prefer to just let `drop` happen in whatever order. Is it important that we explicitly drop the clients and call server.shutdown?

tarpc/src/macros.rs
shaladdle: Use this: https://crates.io/crates/tempdir? I think that generates random directories, so you can do `Tempdir::new().unwrap().path().join("file");`

tikue: I don't think tempdir exposes the path? I'll double check.

tarpc/src/transport/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+        }
+    }
+}
+
+/// A cloneable Reader/Writer.
+pub trait Stream: Read + Write + Send + Sized + 'static {
+    /// Clone that can fail.
+    fn try_clone(&self) -> io::Result<Self>;
+    /// Sets a read timeout.
+    fn set_read_timeout(&self, dur: Option<Duration>) -> io::Result<()>;
```
shaladdle: Do these take `Option<Duration>` because `None` means "Unset the timeout"? If so, please add comments saying so.

tikue: It was just because TcpStrean's equivalent timeout methods take an Option.

tikue: But yeah I guess that's the intended semantics. Noted.

tarpc/src/transport/mod.rs
```diff
@@ -0,0 +1,84 @@
+use std::io::{self, Read, Write};
+use std::time::Duration;
+
+/// A factory for creating a listener on a given address.
```
shaladdle: The word address only comes up in the comments, and I think it's confusing. In a vacuum, it doesn't seem obvious to me what the connection is between an address and a transport, or what you even mean by address.

Some ideas
- Remove the word address 
- Rename the trait to `Address`
- Provide more detailed comments about what addresses have to do with this trait. Maybe an example of what address would mean for tcp, which might make it more concrete/understandable.

tikue: Isn't Addr an associated type of Dialer?

tarpc/src/macros.rs
tikue: Hm, I must have been looking at a different crate last time. Tempdir should work.

tarpc/src/macros.rs
tikue: If we don't call `shutdown` the server won't shut down. I guess that's not a big deal since it'll shut down when the process exits? Will change preemptively...

tarpc/src/macros.rs
tikue: Fixed.

tarpc/src/macros.rs
tikue: Changed.

tarpc/src/transport/mod.rs
tikue: Done.

tarpc/src/transport/unix.rs
shaladdle: Maybe like this:

```rust
self.local_addr().map(|addr| {
  match addr.as_pathname() {
    Some(path) => Ok(UnixDialer(path.to_owned())),
    None => Err(io::Error::new(io::ErrorKind::Other, "tried to create unix socket with invalid path")),
  }
})
```

tarpc/src/transport/mod.rs
tikue: Removed the associated type `Addr` from `Dialer` but still haven't resolved the address issue in comments. Is it just a documentation thing, then?

tarpc/src/transport/unix.rs
tikue: Done.

README.md
```diff
@@ -47,7 +47,7 @@ impl HelloService for HelloServer {
 
 fn main() {
     let server_handle = HelloServer.spawn("0.0.0.0:0").unwrap();
-    let client = hello_service::Client::new(server_handle.local_addr()).unwrap();
+    let client = hello_service::Client::new(server_handle.dialer()).unwrap();
```
shaladdle: Maybe we should have the example in the readme just use a specific IP address? Like localhost:10000 or something? Then people don't have to learn about dialers if they don't care about them.

tarpc/src/macros.rs
```diff
@@ -261,7 +261,7 @@ macro_rules! service {
 #[doc(hidden)]
 #[macro_export]
 macro_rules! service_inner {
-    // Pattern for when the next rpc has an implicit unit return type
+// Pattern for when the next rpc has an implicit unit return type
```
shaladdle: Are these whitespace changes intentional? In the diff they look worse.

tikue: I know they're definitely worse but it's cargo fmt and I wanted to start using the hook lol.

README.md
tikue: Alternatively, we could have the server handle have a method that returns a client.

`server_handle.new_client().unwrap();`

Thoughts?

shaladdle: My thoughts were more that the readme example should be the most common simple case. I think it's more common to have a client/server in different processes, rather than in the same one. So, I was thinking that using an address would be more familiar to some people.

A separate issue is that we don't really have a full suite of examples.. Just the benchmark. That should be addressed separately though.

Returning client seems ok too. The one advantage of having a Dialer is that you can send bogus stuff over a stream, which is useful for verifying that we don't misbehave if someone sends junk to the server.

tikue: I'm fine making the example simpler. The handle should have two methods,
one returning Dialer and one creating a new client.

tarpc/src/protocol/server.rs
```diff
@@ -212,29 +224,33 @@ pub trait Serve: Send + Sync + Sized {
...
+                            -> io::Result<ServeHandle<<T::Listener as Listener>::Dialer>>
+        where T: Transport,
+              Self: 'static
     {
-        let listener = try!(TcpListener::bind(&addr));
-        let addr = try!(listener.local_addr());
-        info!("spawn_with_config: spinning up server on {:?}", addr);
+        let listener = try!(transport.bind());
+        let dialer = try!(listener.dialer());
+        info!("spawn_with_config: spinning up server.");
```
shaladdle: Hmm, now that I see this, it kind of sucks to lose the ability to be able to say the address or 'location' of the server. Maybe `Listener` should impl `Display` or something? Or have an `listen_addr` method?

tarpc/src/transport/mod.rs
shaladdle: Yeah just something like "for TCP an address might be an IPV4 address, for unix sockets it is just a file name" or something along those lines.

tarpc/src/transport/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+
+/// A cloneable Reader/Writer.
+pub trait Stream: Read + Write + Send + Sized + 'static {
+    /// Creates a new independently owned handle to the underlying socket.
+    ///
+    /// The returned TcpStream should reference the same stream that this
+    /// object references. Both handles should read and write the same
+    /// stream of data, and options set on one stream should be propagated
+    /// to the other stream.
+    fn try_clone(&self) -> io::Result<Self>;
```
shaladdle: The above comments seem specific to TCP. Can you make them generic?

tarpc/src/transport/tcp.rs
```diff
@@ -0,0 +1,62 @@
+use std::io;
+use std::net::{SocketAddr, TcpListener, TcpStream, ToSocketAddrs};
+use std::time::Duration;
+
+/// A transport for TCP.
+pub struct TcpTransport<A: ToSocketAddrs>(pub A);
+impl<A: ToSocketAddrs> super::Transport for TcpTransport<A> {
```
shaladdle: Add a line in between the struct and the impl?

tarpc/src/transport/tcp.rs
```diff
@@ -0,0 +1,62 @@
...
+        self.set_write_timeout(dur)
+    }
+    fn shutdown(&self) -> io::Result<()> {
+        self.shutdown(::std::net::Shutdown::Both)
+    }
+}
+
+/// Connects to a socket address.
+pub struct TcpDialer<A = SocketAddr>(pub A) where A: ToSocketAddrs;
+impl<A> super::Dialer for TcpDialer<A>
```
shaladdle: Add a line between impl and struct

tarpc/src/transport/tcp.rs
```diff
@@ -0,0 +1,62 @@
...
+    fn bind(&self) -> io::Result<TcpListener> {
+        TcpListener::bind(self)
+    }
+}
+
+impl super::Listener for TcpListener {
+    type Dialer = TcpDialer<SocketAddr>;
+    type Stream = TcpStream;
+    fn accept(&self) -> io::Result<TcpStream> {
+        self.accept().map(|(stream, _)| stream)
```
shaladdle: Did you think about converting the second element of the tuple to a `Dialer` or something? It seems conceivable that we might want this functionality later.

tarpc/src/transport/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+
+/// Iterates over incoming connections.
+pub struct Incoming<'a, L: Listener + ?Sized + 'a> {
+    listener: &'a L,
+}
+
+impl<'a, L: Listener> Iterator for Incoming<'a, L> {
+    type Item = io::Result<L::Stream>;
+
+    fn next(&mut self) -> Option<Self::Item> {
```
shaladdle: The whitespace style seems inconsistent.. Can you make sure all the new files in this PR follow one way or another? I'd personally prefer a line between items inside impls.

tarpc/src/transport/unix.rs
```diff
@@ -0,0 +1,58 @@
...
+    fn set_read_timeout(&self, timeout: Option<Duration>) -> io::Result<()> {
+        self.set_read_timeout(timeout)
+    }
+    fn set_write_timeout(&self, timeout: Option<Duration>) -> io::Result<()> {
+        self.set_write_timeout(timeout)
+    }
+    fn shutdown(&self) -> io::Result<()> {
+        self.shutdown(::std::net::Shutdown::Both)
+    }
+}
```
shaladdle: Please add a test for this transport.

tarpc/src/protocol/server.rs
tikue: Let's just wait for specialization to land on stable, IMO.

tarpc/src/transport/tcp.rs
tikue: I'll try but it might be from cargo fmt.

tarpc/src/transport/tcp.rs
tikue: The dialer is available on the listener though.

shaladdle: `accept` returns `(TcpStream, SocketAddr)`. The `SocketAddr` is the address of the client, not the address the server is listening on.

README.md
shaladdle: Sounds fine

tarpc/src/transport/tcp.rs
shaladdle: I don't think so. Line 19 on unix.rs has a line between them.

tarpc/src/transport/tcp.rs
tikue: Ah, I see. Yeah, I'm fine making it return (Listener, Dialer).

tikue: The problem is that unix_socket's SocketAddr might be abstract, in which case it doesn't have a path, and thus can't be used as a `Dialer` because `UnixStream::connect` requires a path.

tarpc/src/transport/mod.rs
tikue: Expanded the explanation.

tarpc/src/transport/mod.rs
tikue: How about lines between items in impls only when there aren't comments separating them?

tarpc/src/transport/unix.rs
tikue: Is `async_try_clone_unix` in the `macros` module not enough? I kind of wanted to avoid a combinatorial explosion of tests.

tikue:
All comments addressed. Ready for re-review, I think.

PR #30, opened 2016-04-02
Update serde to 0.7

tikue:
And bincode to 0.5, which is compatible with serde 0.7.


tikue:
There are some random changes for which `cargo fmt` is to blame.

tikue:
How did coverage decrease by so much?! :(

PR #37, opened 2016-05-09
Async via MIO.

### LINKED ISSUE #38: Feature request: more transports (was: stdin/stdout transport for IPC)
The spawner process would act as the server and its children as clients. The child process's stdin and stdout can be treated like a socket. The child process could initiate the connection in its main.

I would use that to build a plugin system that allows live load/reload/unload-ing the plugins.

tikue:
Noteworthy bits:

* Not noticeably faster when working with a single client. The main benefit is multiplexing many clients/services on a single event loop.
* Removed most of the transport abstraction in favor of two enums, `Stream` and `Listener`.
* Provides 3 clients now: `Client` which is callback based, `BlockingClient`, and `FutureClient`.
* Two servers: `BlockingServer` which spawns a thread per rpc, and `Server` which is nonblocking.
* I was able to remove the `*client_methods!` macros by ditching the `Reply` enum. It wasn't required by the protocol since the client always knows what type to expect. We've discussed changing this in the future to return a standard `protocol::Reply` enum, which covers cases like `UnexpectedRequest` (if it couldn't be deserialized...etc.).

-------------------------
# Other changes
* Added `RpcError` and `RpcErrorCode`.
* Changed `BlockingService` trait to have return types of `Result<T, tarpc::RpcError>`
* Changed `Ctx` et al to accept a return value of `Result<T, Into<tarpc::RpcError>>`.
  * This means you can give it `tarpc::Error` and it will convert it for you.
    * This means you can chain the result of one service into the return value of a client request.
* Added an `RpcError` variant of `tarpc::Error`. tarpc uses this to transform `tarpc::RpcError` into `tarpc::Error` behind the scenes for the client, to un-nest the result return type so that it isn't `tarpc::Result<tarpc::RpcResult<T>>`.
* Added server-side-deserialization-error handling. These are converted into `RpcError`s with error code `WrongService`.
* Created two global event loops, one for client and one for server. By default, new clients and servers use the global event loop.
* Removed the `shutdown` methods on the client and server handles. Event loops now can only be shutdown through the registries.
* Renamed `Server::spawn` => `Server::listen` because it doesn't spawn a thread, and generally has a very small resource footprint.
* Renamed `Client::spawn` => `Client::connect` for similar reasons.
* Other renames:
  * `Client` => `AsyncClient`
  * `BlockingClient` => `SyncClient`
  * `Server` => `AsyncServer`
  * `BlockingServer` => `SyncServer`
* Added a `Debug` variant to client and server `Dispatcher` `Action`s and `fn debug` to the registries. This makes it easy to get information about the running dispatchers. Currently only returns the number of handlers running.
* Moved client `fn`s `connect` and `register` into the trait `protocol::client::Client`. This prevents name collisions with rpcs.
* Moved service `fn`s `listen` and `register` into a trait `ServiceExt` that's expanded by the `service!` macro, for the same reason as the previous bullet.
* Moved `__AsyncServer` into `fn register` of `ServiceExt` to prevent unlikely collisions.
* Moved `Ctx` and `SendCtx` out of the macro and into `protocol::server`. Instead the macro
  creates a `Ctx` extension trait that `protocol::server::Ctx/SendCtx` `impl`.
* Replaced `Ctx::rpc` finisher methods with a single `Ctx::reply` method, and made `Ctx` generic
  over the reply type, i.e. the ctx arg type to the service fns is now `Ctx<Reply>`.
* Added a `CachedPool` for use by `SyncService`. It is a growable thread pool whose idle threads expire on a configurable timer.
  * Can also configure the minimum number of threads to keep around (so that it doesn't always expire to 0) and the maximum number of threads to ever spawn.
* Support for more transports.
  * Unix sockets
  * Pipes (stdin/stdout)

Closes #38


Cargo.toml
```diff
@@ -12,12 +12,23 @@ description = "An RPC framework for Rust with a focus on ease of use."
...
 [dependencies]
 bincode = "0.5"
+bytes = "0.3"
+byteorder = "0.5"
 log = "0.3"
+mio = "0.5"
+quick-error = "1.0"
 scoped-pool = "0.1"
 serde = "0.7"
+env_logger = "*"
```
shaladdle: Specify version?

tikue: Done.

src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+
+use mio::NotifyError;
+use std::io;
+use std::sync::mpsc;
+
+quick_error! {
+    /// All errors that can occur during the use of tarpc.
+    #[derive(Debug)]
+    pub enum Error {
+        /// No address found for the specified address.
```
shaladdle: I don't understand this comment. Please expand on it.

src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+        Io(err: io::Error) {
+            from()
+            description(err.description())
+        }
+        /// Error in receiving a value, typically from an event loop that may have shutdown.
+        Rx(err: mpsc::RecvError) {
+            from()
+            description(err.description())
+        }
+        /// Error in serializing, either on client or server.
```
shaladdle: This comment seems to be swapped with the one for the `Serialize` variant.

tikue: Done.

src/lib.rs
tikue: Done.

tikue:
Coveralls pls

tailhook:
Consider building on top of `rotor` instead. It would allow integrating multiple things into the same event loop. E.g. combine it with hyper. FYI, async branch of hyper runs on top of `rotor` too. We also working on async DNS which tarpc could also benefit from.

tikue:
@tailhook Rotor allows you to integrate multiple things at compile-time. I've built this in such a way that you can register new tarpc services dynamically on a global event loop, which doesn't need to have knowledge of the specifics of the services.

tikue:
@tailhook sorry, looks like I jumped the gun -- rotor actually supports adding machines dynamically too? I'll have to look into it more.

tailhook:
@tikue sure. Feel free to file issues if you have any difficulties. We don't have much docs, and real questions from users will encourage me to document right things :)

tikue:
Tests hang on windows.

tikue:
https://travis-ci.org/google/tarpc/jobs/135960990 

> /Users/travis/build.sh: line 45: travis-cargo: command not found

No idea what's going on there.

README.md
```diff
@@ -6,18 +6,22 @@
...
 
 ## What is an RPC framework?
-"RPC" stands for "Remote Procedure Call," a function call where the work of producing the return
-value is being done somewhere else. When an rpc function is invoked, behind the scenes the function
-contacts some other process somewhere and asks them to compute the function instead. The original
-function then returns the value produced by that other server.
+"RPC" stands for "Remote Procedure Call," a function call where the work of
+producing the return value is being done somewhere else. When an rpc function is
+invoked, behind the scenes the function contacts some other process somewhere
+and asks them to compute the function instead. The original function then
```
shaladdle: Nit: s/compute/evaluate

README.md
```diff
@@ -6,18 +6,22 @@
...
+returns the value produced by that other server.
```
shaladdle: s/that other server/the other process

README.md
```diff
@@ -28,57 +32,68 @@ tarpc = "0.5.0"
...
 ```
 
-The `service!` macro expands to a collection of items that collectively form an rpc service. In the
-above example, the macro is called within the `hello_service` module. This module will contain a
-`Client` (and `AsyncClient`) type, and a `Service` trait. The trait provides default `fn`s for
-starting the service: `spawn` and `spawn_with_config`, which start the service listening over an
-arbitrary transport. A `Client` (or `AsyncClient`) can connect to such a service. These generated
-types make it easy and ergonomic to write servers without dealing with sockets or serialization
-directly. See the tarpc_examples package for more sophisticated examples.
+The `service!` macro expands to a collection of items that collectively form an
```
shaladdle: Remove "collectively", seems redundant with "collection".

README.md
```diff
@@ -28,57 +32,68 @@ tarpc = "0.5.0"
...
-- Attributes can be specified on rpc methods. These will be included on both the `Service` trait
-  methods as well as on the `Client`'s stub methods.
+- Backed by an mio `EventLoop`, protecting services (including `SyncService`s)
+  from slowloris attacks.
+- Run any number of clients on a single client event loop.
+- Run any number of services on a single service event loop.
+- Configure clients and services to run on a custom event loop, defaulting to
+  the global event loop.
+- Any type that `impl`s `serde`'s `Serialize` and `Deserialize` can be used in
+  the rpc signatures.
```
shaladdle: Nit: remove "the"

README.md
```diff
@@ -28,57 +32,68 @@ tarpc = "0.5.0"
...
+- Configurable server rate limiting.
+- Automatic client retries with exponential backoff when server is busy.
 - Concurrent requests from a single client.
-- Any type that `impl`s `serde`'s `Serialize` and `Deserialize` can be used in the rpc signatures.
-- Attributes can be specified on rpc methods. These will be included on both the `Service` trait
-  methods as well as on the `Client`'s stub methods.
+- Backed by an mio `EventLoop`, protecting services (including `SyncService`s)
+  from slowloris attacks.
+- Run any number of clients on a single client event loop.
+- Run any number of services on a single service event loop.
```
shaladdle: Can I mix clients and services on one event loop?

README.md
```diff
@@ -28,57 +32,68 @@ tarpc = "0.5.0"
...
+- Configure clients and services to run on a custom event loop, defaulting to
+  the global event loop.
+- Any type that `impl`s `serde`'s `Serialize` and `Deserialize` can be used in
+  the rpc signatures.
+- Attributes can be specified on rpc methods. These will be included on both the
+  services' trait methods as well as on the clients' stub methods.
 - Just like regular fns, the return type can be left off when it's `-> ()`.
```
shaladdle: I guess I'd lean towards not listing these two, since intuitively I would expect the macro to handle most of the same syntax as rust fns.

examples/cached_pool.rs
```diff
@@ -0,0 +1,46 @@
...
+    let pools = &[CachedPool::new(Config::max_idle(Duration::from_secs(5))),
+                  CachedPool::new(Config::max_idle(Duration::from_millis(500)))];
+    for _ in 0..15 {
+        for pool in pools {
+            pool.execute(move || {
+                    thread::sleep(Duration::from_secs(5));
+                })
+                .unwrap();
+        }
+        info!("{:?}",
```
shaladdle: I confess to not have any idea what the return type should be here, or what this is suppoed to do.. Can you please adda comment?

examples/concurrency.rs
```diff
@@ -0,0 +1,112 @@
...
+
+fn run_once(clients: &[FutureClient], concurrency: u32, print: bool) {
+    let _ = env_logger::init();
+
+    let start = Instant::now();
+    let futures: Vec<_> = 
+        clients.iter()
+               .cycle()
+               .take(concurrency as usize)
+               .map(|client| client.read(&CHUNK_SIZE, &SystemTime::now().serializable()))
```
shaladdle: Nit: is there a reason you don't just do this:

```
.map(|client| (client.read(&CHUNK_SIZE), &SystemTime::now().serializable()))
```

And then you can just do now - start when you compute latencies. Why is the time sent with the rpc?

examples/pipe.rs
```diff
@@ -0,0 +1,43 @@
...
+//
+// Licensed under the MIT License, <LICENSE or http://opensource.org/licenses/MIT>.
+// This file may not be copied, modified, or distributed except according to those terms.
+
+#![feature(default_type_parameter_fallback, try_from)]
+extern crate mio;
+#[macro_use]
+extern crate tarpc;
+
+#[cfg(unix)]
```
shaladdle: Nit: Is there some way to just exclude this whole file from the travis build on non-unix instead of throwing cfg(unix) all over the place?

examples/rate_limited.rs
```diff
@@ -0,0 +1,65 @@
...
+            format!("{} - {} - {}",
+                Local::now(),
+                record.level(),
+                record.args()
+            )
+        };
+
+    let mut builder = LogBuilder::new();
+    builder.format(format);
+    if env::var("RUST_LOG").is_ok() {
```
shaladdle: Nit:

```rust
if let Ok(s) = env::var("RUST_LOG") {
  builder.parse(&s);
}
```

examples/readme.rs
```diff
@@ -0,0 +1,36 @@
...
+    }
+}
+
+fn main() {
+    extern crate bincode;
+    env_logger::init().unwrap();
+    println!("{:?}", bincode::serde::serialize(&"hi".to_string(), bincode::SizeLimit::Infinite).unwrap());
+    let addr = "localhost:10000";
+    let _server = HelloServer.listen(addr).unwrap();
+    thread::park();
```
shaladdle: What's this do?

src/cached_pool/mod.rs
```diff
@@ -0,0 +1,640 @@
...
+
+    /// Creates a new `Config` with the specified minimum number of threads, and all other fields
+    /// set to defaults.
+    #[inline]
+    pub fn min_threads(min_threads: u32) -> Config {
+        Config { min_threads: min_threads, ..Config::default() }
+    }
+
+    #[inline]
+    fn max_idle_ms(&self) -> u64 {
```
shaladdle: Is the unwrap necessary here? If we're going to unwrap here, maybe add an assert or something when specifying max idle time to make sure it's safe to call max_idle_ms? Also, this function seems to only be used in one place, maybe we can just remove it.

src/cached_pool/mod.rs
```diff
@@ -0,0 +1,640 @@
...
+            tx: tx,
+            count: Some(Arc::new(())),
+        }
+    }
+
+    /// Submit a new task to a thread.
+    ///
+    /// Fails if all threads are busy with tasks, and the thread pool
+    /// is running its maximum configured number of threads.
+    fn execute<F>(&self, pool_id: PoolId, f: F) -> Result<(), Box<Task + Send>>
```
shaladdle: There's quite a bit of unwrapping here.. Are these all things that should never happen? I think that it would be good if we could have a comment for each unwrap that isn't obviously ok (eg mutex lock and thread join unwraps are pretty obvious to me)

README.md
tikue: No

examples/cached_pool.rs
tikue: Gonna push back a bit on this, because I don't think we should comment all iterator chained methods. Can you elaborate on what's confusing you so I have a better idea of what needs commenting?

examples/concurrency.rs
tikue: It's just part of the example. IMO it makes for a more interesting example to send the time as part of the rpc.

examples/pipe.rs
tikue: I'd also like to know this! I agree it's super ugly.

examples/rate_limited.rs
tikue: Oh wow, I don't think I'd ever do this myself. I think I copied this part wholesale from a google search for "env_logger custom formatting"

examples/readme.rs
tikue: The thread won't be woken up until another thread calls unpark. We probably should have a "block forever" method on the server handle.

src/cached_pool/mod.rs
tikue: The ergonomics loss for a situation that will never happen in practice makes me uncomfortable. I'd rather just document that restriction. It's the same thing mio does.

tikue: Added a check on `fn max_idle` but the user can still create the struct from the public fields.

tikue:
I authored these commits

src/cached_pool/mod.rs
tikue: `mio::Sender::send` can fail if mio's backing queue (which is fixed-size) is full. I don't want to expose this in the API, and I intend to replace mio's backing queue with a custom queue that grows indefinitely, once such an option is available. At that point, it will be safe to call unwrap. Until that point, these can technically panic, but it's unlikely.

tikue: Channel sends and receives are the form of communication used ubiquitously in the thread pool code, and it'd be pretty tiresome to document/read the same comment over and over. If anything, perhaps a note at the top of the module about why the unwraps are safe (once they are safe).

tikue:
Closing in favor of #44

PR #49, opened 2016-09-05
Remove two unused error variants of tarpc::Error.

tikue:
1. `NoAddressFound` - we return `io::Error` from `fn connect` and `fn listen` now, so this variant of `tarpc::Error` is no longer needed.
2. `ReplyCanceled` - I don't remember how this was previously used, but now any error type returned by an rpc impl will be encoded in the `App` variant.

This change is nice because now there are only three classes of errors we expose:

1. I/O
2. Serialization
3. Application-specific

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -3,7 +3,7 @@
 // Licensed under the MIT License, <LICENSE or http://opensource.org/licenses/MIT>.
 // This file may not be copied, modified, or distributed except according to those terms.
 
-use {bincode, futures};
+use bincode;
 use std::{fmt, io};
 use std::error::Error as StdError;
 use tokio_proto::pipeline;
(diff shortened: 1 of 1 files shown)


src/errors.rs
```diff
@@ -39,16 +34,16 @@ pub enum Error<E>
     ///
     /// Typically this indicates a faulty implementation of `serde::Serialize`.
     ServerSerialize(String),
-    /// The server canceled the response before it was completed.
-    ReplyCanceled,
     /// The server was unable to reply to the rpc for some reason.
+    ///
+    /// This is a service-specific error. It is individually specified in the
+    /// `service!` macro for each rpc.
```
shaladdle: > It is individually specified in the `service!` macro for each rpc.

I think it's more accurate to say it's specified in the implementation of the method, rather than specified in the `service!` macro unless I'm missing something.

tikue: It is returned in the implementation, but the type is specified in the macro invocation. How about, "The type is individually specified in the `service!` macro for each rpc."

shaladdle: Good point, that sounds good.

tikue: Done!

shaladdle:
LGTM, merging

shaladdle:
(When checks complete)

PR #55, opened 2016-09-27
Replace handrolled transport with higher-level `Framed` construct

tikue:


tikue:
I authored these commits.

shaladdle:
I authored these commits.

src/framed.rs
```diff
@@ -0,0 +1,176 @@
...
+use serde;
+use std::io::{self, Cursor};
+use std::marker::PhantomData;
+use std::mem;
+use tokio_core::easy::{self, EasyBuf, EasyFramed};
+use tokio_core::io::{FramedIo, Io};
+use tokio_proto::multiplex::{self, RequestId};
+use util::Never;
+
+/// Handles the IO of tarpc messages.
```
shaladdle: How is it different from `EasyFramed`?

src/framed.rs
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    fn flush(&mut self) -> Poll<(), io::Error> {
+        self.inner.flush()
+    }
+}
+
+struct Parser<T> {
+    state: ParserState,
+    _phantom_data: PhantomData<T>,
```
shaladdle: What's `T`? Can we have a comment about why we need `PhantomData`?

src/framed.rs
```diff
@@ -0,0 +1,176 @@
...
+{
+    type In = (RequestId, T);
+
+    fn serialize(&mut self, (id, message): Self::In, buf: &mut Vec<u8>) {
+        buf.write_u64::<BigEndian>(id);
+        buf.write_u64::<BigEndian>(bincode::serialized_size(&message));
+        bincode::serialize_into(buf,
+                                &message,
+                                SizeLimit::Infinite)
+                 // TODO(tikue): handle err
```
shaladdle: Anything resolving blocking this TODO?

src/macros.rs
```diff
@@ -329,51 +325,54 @@ macro_rules! service {
             rpc $fn_name( $( $arg : $in_ ),* ) -> $out | $error;
         }
     };
-    // Pattern for when all return types have been expanded
+// Pattern for when all return types have been expanded
```
shaladdle: :( indents look worse. Is this rustfmt's doing?

src/macros.rs
```diff
@@ -533,14 +528,23 @@ macro_rules! service {
...
                 where L: ::std::net::ToSocketAddrs
             {
+                let addr = if let ::std::option::Option::Some(a) =
+                    ::std::iter::Iterator::next(
+                        &mut ::std::net::ToSocketAddrs::to_socket_addrs(&addr)?) {
+                    a
+                } else {
+                    return Err(::std::io::Error::new(::std::io::ErrorKind::AddrNotAvailable,
+                               "`ToSocketAddrs::to_socket_addrs` returned an empty iterator."));
+                };
```
shaladdle: Can we pull these lines into a helper that's not in a macro?

examples/pubsub.rs
```diff
@@ -121,14 +117,20 @@ impl publisher::FutureService for Publisher {
 
 fn main() {
     let _ = env_logger::init();
-    let publisher = Publisher::new().listen("localhost:0").wait().unwrap();
-    let publisher = publisher::SyncClient::connect(publisher.local_addr()).unwrap();
-    let _subscriber1 = Subscriber::new(0, publisher.clone());
-    let _subscriber2 = Subscriber::new(1, publisher.clone());
+    let publisher = Publisher::new().listen("localhost:0".first_socket_addr()).wait().unwrap();
+    let publisher_addr = publisher.local_addr();
+    let publisher = publisher::SyncClient::connect(publisher_addr).unwrap();
```
tikue: Done.

src/client.rs
```diff
@@ -3,76 +3,117 @@
...
 use std::fmt;
 use std::io;
-use tokio_proto::pipeline;
+use tokio_core::net::TcpStream;
+use tokio_core::reactor;
+use tokio_proto::easy::{EasyClient, EasyResponse, multiplex};
 use tokio_service::Service;
-use util::Never;
 
 /// A client `Service` that writes and reads bytes.
```
tikue: I think EasyClient is probably a wrapper around a tokio service. `EasyClient` is an implementation detail though, so we probably don't need to mention that. I think the important thing is that `Client` impls `tokio_service::Service`. I clarified a bit.

src/client.rs
```diff
@@ -3,76 +3,117 @@
...
+              E: Deserialize + Send + 'static
+    {
+        Client {
+            inner: multiplex::connect(Framed::new(tcp), handle),
+        }
+    }
+
+    fn map_err(resp: WireResponse<Resp, E>) -> Result<Resp, ::Error<E>> {
+        resp.map(|r| r.map_err(::Error::from))
+            .map_err(::Error::ClientDeserialize)
```
tikue: `E` is not guaranteed to be `SerializeError`. `E` is just `E`, but `map_err` is being called on `WireResponse`, whose error type is `DeserializeError`:

```rust
type WireResponse<Resp, E> = Result<Result<Resp, WireError<E>>, DeserializeError>;
```

src/client.rs
```diff
@@ -81,34 +122,83 @@ pub mod future {
...
+        type Item = Client<Req, Resp, E>;
+        type Error = io::Error;
 
-        /// Starts an event loop on a thread and registers a new client
-        /// connected to the given address.
-        fn connect(addr: &SocketAddr) -> ClientFuture {
+        fn poll(&mut self) -> futures::Poll<Self::Item, Self::Error> {
+            match self.inner.poll()? {
+                Async::Ready(client) => Ok(Async::Ready(client)),
+                Async::NotReady => Ok(Async::NotReady),
```
tikue: Hahaha, this was just a bad copy-paste.

src/framed.rs
tikue: Presumably we might want to change the implementation of `Framed` along the way, so I didn't want to expose the fact that we're currently using `EasyFramed`.

src/framed.rs
tikue: `T` is the type being parsed. Added a comment.

src/macros.rs
tikue: Yeah...

src/macros.rs
tikue: Good idea. Done.

src/client.rs
```diff
@@ -3,55 +3,74 @@
 // Licensed under the MIT License, <LICENSE or http://opensource.org/licenses/MIT>.
 // This file may not be copied, modified, or distributed except according to those terms.
 
-use Packet;
-use futures::{Async, BoxFuture};
+use {WireError, tokio_proto as proto};
```
tikue: imports are relative from the crate root; `WireError` wasn't already in scope. It's similar to how if you don't say `use std;` then you have to prefix paths with `::std::` rather than `std::`.

src/client.rs
```diff
@@ -3,55 +3,74 @@
...
+        self.inner
+            .call(proto::Message::WithoutBody(request))
+            .map(Self::map_err)
     }
 }
 
-impl fmt::Debug for Client {
+impl<Req, Resp, E> Client<Req, Resp, E> {
+    fn map_err(resp: WireResponse<Resp, E>) -> Result<Resp, ::Error<E>> {
+        resp.map(|r| r.map_err(::Error::from))
```
tikue: `Error` is defined in the `errors` module and reexported at the crate root.

src/client.rs
```diff
@@ -81,25 +100,38 @@ pub mod future {
...
+                                let c =
+                                    multiplex::connect(move || {
+                                                           Ok(Framed::new(tcp.borrow_mut().take().unwrap()))
+                                                       },
+                                                       &handle2);
+                                tx.complete(Ok(Client { inner: c }));
+                            }
+                            Err(e) => {
+                                tx.complete(Err(e));
+                            }
```
tikue: This was changed while addressing another comment.

src/framed.rs
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    fn flush(&mut self) -> io::Result<Async<()>> {
+        self.inner.flush()
+    }
+}
+
+struct Parser<T> {
+    state: ParserState,
+    _phantom_data: PhantomData<T>
```
tikue: Done.

src/framed.rs
tikue: Did you mean to say "anything blocking resolving this TODO?"?

src/framed.rs
```diff
@@ -0,0 +1,244 @@
...
+use bytes::{Buf, MutBuf};
+use bytes::buf::{BlockBuf, BlockBufCursor};
+use std::{cmp, io, mem};
+use std::marker::PhantomData;
+use util::Never;
+use tokio_core::io::{FramedIo, Io};
+use tokio_proto::{self as proto, multiplex};
+
+/// Handles the IO of tarpc messages.
+pub struct Framed<I, In, Out> {
```
shaladdle: Why would you have to do that?

src/framed.rs
shaladdle: Ah I see it's public. I was thinking more about comments to help those reading the implementation.

src/macros.rs
shaladdle: We can leave them as bad if you want. There's lots of merit to being able to rustfmt everything. Do you know if there's an issue open for this?

src/framed.rs
tikue: Otherwise the type parameter would be shadowing the trait

src/macros.rs
tikue: Yeah let's leave them. I tried to keep them nice for a while, which was a pretty frustrating experience.

tikue: It looks like a new version hasn't been released in like 2 months? Not sure why.

shaladdle: Up to you. Leaving it until they make a release is fine with me.

src/framed.rs
tikue: I guess the main difference is that EasyFramed's type parameters specify the parser and serializer to use -- and by those types, implicitly specifies the types being parsed and serialized -- whereas tarpc::Framed's type parameters are the actual types being parsed and serialized. Not sure if it's worth explaining that in a comment?

src/macros.rs
tikue: The latest rustfmt on git didn't seem to fix this. Oh well...

PR #61, opened 2016-12-05
Track the changes to tokio-proto/master

### LINKED ISSUE #60: Breakage caused by recent changes to tokio
As far as I can tell there are three changes: removal of poll_ready from the Service trait in tokio-service and tokio-proto; rename and changes to tokio_core::easy; changes to the ways that NewService can be implemented.

I was able to work around the first two by removing implementations of poll_ready and pulling a older copy of easy.rs into tarpc. I am struggling to understand how to respond to the NewService change though. This problem shows up when I try to build readme_future.rs.

tikue:
Fixes #60

### DIFF
--- a/examples/throughput.rs
+++ b/examples/throughput.rs
@@ -52,8 +52,8 @@ impl FutureService for Server {
 const CHUNK_SIZE: u32 = 1 << 19;
 
 fn bench_tarpc(target: u64) {
-    let handle = Server.listen("localhost:0".first_socket_addr()).wait().unwrap();
-    let client = SyncClient::connect(handle.local_addr()).unwrap();
+    let addr = Server.listen("localhost:0".first_socket_addr()).wait().unwrap();
+    let client = SyncClient::connect(&addr).unwrap();
     let start = time::Instant::now();
     let mut nread = 0;
     while nread < target {
--- a/src/framed.rs
+++ b/src/framed.rs
@@ -18,7 +18,7 @@ use util::{Debugger, Never};
 pub type Frame<T> = multiplex::Frame<T, Never, io::Error>;
 
 
-// `T` is the type that `Codec` parses.
+// `Req` is the type that `Codec` parses. `Resp` is the type it serializes.
 pub struct Codec<Req, Resp> {
     state: CodecState,
     _phantom_data: PhantomData<(Req, Resp)>,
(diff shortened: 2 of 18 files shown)


src/framed.rs
```diff
@@ -3,170 +3,172 @@
...
-    fn flush(&mut self) -> Poll<(), io::Error> {
-        self.inner.flush()
-    }
-}
 
-// `T` is the type that `Parser` parses.
-struct Parser<T> {
-    state: ParserState,
-    _phantom_data: PhantomData<T>,
+// `T` is the type that `Codec` parses.
```
shaladdle: This looks stale. `T` isn't a thing in this struct defn anymore.

src/server.rs
```diff
@@ -29,35 +29,63 @@ pub fn listen<S, Req, Resp, E>(addr: SocketAddr, new_service: S) -> ListenFuture
...
-               .map_err(|()| panic!("What do we do here")))
+    let listener = listener(&addr, &handle)?;
+    let addr = listener.local_addr()?;
+
+    let handle2 = handle.clone();
+    let server = listener.incoming().for_each(move |(socket, _)| {
+        Proto::new().bind_server(&handle2, socket, new_service.new_service()?);
+
+        Ok(())
+    }).map_err(|e| error!("While processing incoming connections: {}", e));
```
tikue: I've never triggered it in testing but it can definitely happen. I guess if there's a problem accepting a request

src/server.rs
```diff
@@ -29,35 +29,63 @@ pub fn listen<S, Req, Resp, E>(addr: SocketAddr, new_service: S) -> ListenFuture
...
+
+fn listener(addr: &SocketAddr,
+            handle: &Handle) -> io::Result<TcpListener> {
+    match *addr {
+        SocketAddr::V4(_) => net2::TcpBuilder::new_v4(),
+        SocketAddr::V6(_) => net2::TcpBuilder::new_v6()
+    }?
+    .reuse_address(true)?
+    .bind(addr)?
+    .listen(1024)
```
tikue: It's the pending connection backlog. Made it a const.

examples/throughput.rs
```diff
@@ -52,8 +52,8 @@ impl FutureService for Server {
 const CHUNK_SIZE: u32 = 1 << 19;
 
 fn bench_tarpc(target: u64) {
-    let handle = Server.listen("localhost:0".first_socket_addr()).wait().unwrap();
```
tikue: Yeah we should return a handle that has a way to gracefully shut down the service. I opened an #62.

src/framed.rs
tikue: Done.

PR #76, opened 2017-01-13
fix(futures): `Either` was only added in 0.1.7

compressed:


tikue:
With `futures = 0.1`, `cargo update` already does the right thing and will update to any version less than 1.0.

compressed:
Yes, but imagine you had another crate in your project that was depending on `futures = "=0.1.6"` then cargo will select that as version for tarpc but then tarpc won't compile because `0.1.6` isn't compatible.

tikue:
Good point!

PR #78, opened 2017-01-17
Simplify Never impls: return the inner uninhabited type for fns taking &self.

tikue:


shaladdle:
I'm ok with my commits being included in this PR.

PR #79, opened 2017-01-17
Make private a couple items that no longer need to be public.

tikue:


shaladdle:
I wish you were smarter, CLA bot. I definitely approve my commit being included, for the record.

PR #81, opened 2017-01-19
Add TLS support

### LINKED ISSUE #77: Feature:  SSL/TLS Support
Any plans to add support for SSL/TLS?

compressed:
So here we go... TLS support take two!

The `Options` change and recent refactoring made this change much easier to make. I didn't need to modify the macro at all. I was able to re-use all the tests as well with some refactoring. I added the `t!` macro there in testing (it's common in a few other libraries), mainly because it provides better errors (line numbers) on mac. I can change those to unwraps though or change everything to `t!` if you want.

I think most of the changes are straight-forward. The only tricky part I ran into was that `Options` is not `Send`, so I needed to pull out the related TLS part of `Options` on both the client and server in order to make the compiler happy.

Note: I did a merge because the changes in both this codebase and my branch were so much that the rebase was turning out to be quite ugly. I believe you can do a merge and squash and GitHub though.

Fix #77


tikue:
Thanks again for doing this! Hopefully I'll get through it all by the end of the weekend.

src/client.rs
```diff
@@ -174,15 +234,49 @@ pub mod future {
...
-        extern "rust-call" fn call_once(self, (tcp,): (TcpStream,)) -> Client<Req, Resp, E> {
-            Client::new(Proto::new().bind_client(&self.0, tcp))
+        extern "rust-call" fn call_once(self, (stream,): (I,)) -> Self::Output {
+            Client::new(Proto::new().bind_client(&self.0, stream.into()))
+        }
+    }
+
+    /// Provides the connection Fn impl for Tls
+    #[doc(hidden)]
+    pub struct ConnectFn {
```
tikue: I could be mistaken, but I have a feeling this can be private?

src/server.rs
```diff
@@ -50,34 +75,48 @@ pub fn listen<S, Req, Resp, E>(new_service: S, addr: SocketAddr, options: Option
...
+                inner: future::Either::B(future::ok(listen_with(new_service,
+                                                                addr,
+                                                                handle,
+                                                                acceptor))),
             }
         }
     }
 }
 /// Spawns a service that binds to the given address using the given handle.
 #[doc(hidden)]
```
tikue: No longer needed.

src/client.rs
```diff
@@ -19,8 +21,34 @@ use tokio_service::Service;
...
+type BindClient<Req, Resp, E> = <Proto<Req, Result<Resp, WireError<E>>> as
+                                ProtoBindClient<Multiplex, StreamType>>::BindClient;
+
+/// TLS-specific functionality
+#[cfg(feature = "tls")]
+pub mod tls {
+    use native_tls::TlsConnector;
+
+    /// TLS context
+    pub struct TlsClientContext {
```
tikue: I don't feel super strongly about this, but since it's already in the `client::tls` module, we might consider just calling it `Context` and people can alias it if they so choose. Thoughts? Maybe the `tls` module should be top-level and it should be `tls::client::Context`?

cc @shaladdle

src/client.rs
```diff
@@ -19,8 +21,34 @@ use tokio_service::Service;
...
+        /// Domain to connect to
+        pub domain: String,
+        /// TLS connector
+        pub tls_connector: TlsConnector,
+    }
+
+    impl TlsClientContext {
+        /// Try to make a new `TlsClientContext`, providing the domain the client will
+        /// connect to.
+        pub fn try_new<S: Into<String>>(domain: S)
```
tikue: I think it's fine to call this `new`. There's kind of a precedent to having `new` fns return a result, such as `tokio_core::reactor::Core::new()`

src/client.rs
```diff
@@ -19,8 +21,34 @@ use tokio_service::Service;
...
+                                        -> Result<TlsClientContext, ::native_tls::Error> {
+            Ok(TlsClientContext {
+                domain: domain.into(),
+                tls_connector: TlsConnector::builder()?.build()?,
```
tikue: What else can the builder configure? Should we expose a builder for the inner builder? Should there be a separate `fn from_builder(builder: TlsConnectorBuilder) -> Result<Self, ::native_tls::Error>`?

src/client.rs
```diff
@@ -19,8 +21,34 @@ use tokio_service::Service;
...
+        pub domain: String,
+        /// TLS connector
+        pub tls_connector: TlsConnector,
+    }
+
+    impl TlsClientContext {
+        /// Try to make a new `TlsClientContext`, providing the domain the client will
+        /// connect to.
+        pub fn try_new<S: Into<String>>(domain: S)
+                                        -> Result<TlsClientContext, ::native_tls::Error> {
```
tikue: nit: prefer `Self` where applicable, e.g. `Result<Self, ::native_tls::Error>`

src/client.rs
```diff
@@ -91,9 +119,12 @@ impl<Req, Resp, E> fmt::Debug for Client<Req, Resp, E>
 }
 
 /// Additional options to configure how the client connects and operates.
-#[derive(Clone, Default)]
+#[cfg_attr(feature = "tls", derive(Default))]
+#[cfg_attr(not(feature = "tls"), derive(Clone, Default))]
```
tikue: Let's just have it not be `Clone`. I don't feel super comfortable with the traits impl'd changing based on which set of features are enabled.

cc @shaladdle

src/client.rs
```diff
@@ -138,11 +183,26 @@ pub mod future {
     pub struct ConnectFuture<Req, Resp, E>
         where Req: Serialize + 'static,
               Resp: Deserialize + 'static,
-              E: Deserialize + 'static,
+              E: Deserialize + 'static
     {
+        #[cfg(not(feature = "tls"))]
+        inner:
+            future::Either<
```
tikue: Can we add a type alias for part of this so we don't have to rewrite the entire thing in both branches of the `cfg`?

src/lib.rs
```diff
@@ -143,3 +195,12 @@ enum Reactor {
     Handle(reactor::Handle),
     Remote(reactor::Remote),
 }
+
+cfg_if! {
+    if #[cfg(feature = "tls")] {
+        extern crate tokio_tls;
+        pub extern crate native_tls;
+
+        pub use client::tls::TlsClientContext;
```
tikue: Related to my other comment, I'm starting to feel more strongly that there should be a `mod tls` at the crate root, within which is a client module.

src/macros.rs
```diff
@@ -688,21 +688,168 @@ mod syntax_test {
...
 #[cfg(test)]
 mod functional_test {
+    use {client, server};
     use futures::{Future, failed};
+    use std::io;
+    use std::net::SocketAddr;
     use util::FirstSocketAddr;
     extern crate env_logger;
 
+    macro_rules! t {
```
tikue: This is basically unwrap on steroids, right? Is there a mnemonic to remember that?

src/lib.rs
```diff
@@ -143,3 +195,12 @@ enum Reactor {
     Handle(reactor::Handle),
     Remote(reactor::Remote),
 }
+
+cfg_if! {
+    if #[cfg(feature = "tls")] {
+        extern crate tokio_tls;
+        pub extern crate native_tls;
```
tikue: Does this have to be public?

src/server.rs
```diff
@@ -17,10 +17,28 @@ use tokio_core::reactor::{self, Handle};
...
+enum Acceptor {
+    Tcp,
+    #[cfg(feature = "tls")]
+    Tls(TlsAcceptor),
+}
+
 /// Additional options to configure how the server operates.
-#[derive(Clone, Default)]
+#[cfg_attr(feature = "tls", derive(Default))]
+#[cfg_attr(not(feature = "tls"), derive(Clone, Default))]
```
tikue: Same comment as above: I'd rather impl the intersection of possible traits for each feature.

src/client.rs
```diff
@@ -108,21 +139,35 @@ impl Options {
...
+    use stream_type::StreamType;
     use super::{Client, Options};
-    use tokio_core::{self, reactor};
-    use tokio_core::net::TcpStream;
+    #[cfg(feature = "tls")]
+    use super::tls::TlsClientContext;
+    use tokio_core::net::{TcpStream, TcpStreamNew};
+    use tokio_core::reactor;
     use tokio_proto::BindClient;
+    #[cfg(feature = "tls")]
```
tikue: Can we use a `cfg_if!` to group all the tls imports together?

.travis.yml
```diff
@@ -21,7 +21,8 @@ before_script:
 
 script:
 - |
-    travis-cargo build && travis-cargo test
+    travis-cargo build && travis-cargo test &&
+    travis-cargo build -- --features tls && travis-cargo test -- --features tls
 
 after_success:
 - travis-cargo coveralls --no-sudo
```
compressed: These run just like `cargo build --features tls`. If you pass `--` to `travis-cargo` then whatever follows is passed to `cargo`.

src/client.rs
compressed: Yeah, I actually was considering that as well. I didn't particularly like putting `Client` in the name, but thought it would be useful if it was available at the root namespace; `::TlsContext` wouldn't be very descriptive.

I like the idea of a root `tls` module.

src/client.rs
compressed: Yup, agreed. Will do.

src/client.rs
compressed: Makes sense, will do.

src/client.rs
compressed: Okay, yeah I wasn't sure if you had some need for `Clone` but I agree it's better if it doesn't change between features.

src/client.rs
compressed: Oh yes, awesome!

Rust was originally complaining about private type in a public interface, but that was with an older version of the `Connect` trait. Making this private seems okay now. I like that it can be private again as well.

src/lib.rs
compressed: Well we need to expose the `TlsAcceptor` type in the builder for the server options. Another option here though instead of just re-exporting the whole crate would be to just re-export the `TlsAcceptor` type itself.

I was also thinking on the `client` side it could take an option `TlsConnector` if people want to override the default. So we could get away with just re-exporting `TlsConnector` and `TlsAcceptor` from `native_tls` instead of the whole crate; that's probably a better idea.

src/lib.rs
compressed: Yes, I like this approach too. I will make those changes.

src/macros.rs
compressed: Ha yes. I'm not sure on the history behind the naming though, other than it is short to type...

README.md
```diff
@@ -133,6 +133,78 @@ fn main() {
...
+     (acceptor, client_cx)
+}
+
+fn main() {
+    let addr = "localhost:10000".first_socket_addr();
+    let mut core = reactor::Core::new().unwrap();
+    let (acceptor, client_cx) = tls_context();
+    HelloServer.listen(addr, server::Options::default()
+                                 .handle(core.handle())
+                                 .tls(acceptor)).wait().unwrap();
```
compressed: Good idea!

src/client.rs
compressed: Yup, will do.

Cargo.toml
```diff
@@ -33,7 +36,12 @@ env_logger = "0.3"
 futures-cpupool = "0.1"
 clap = "2.0"
 
+[target.'cfg(target_os = "macos")'.dev-dependencies]
```
compressed: This line will make this dependency apply only for osx. For linux, it internally uses the openssl crate, but we can make required data for testing purposes without needing to bring in the openssl crate.

For windows, I didn't write the test case since I don't actually have access to a windows machine and it seemed a little complicated based on the tests I read in the `native-tls` and `tokio-tls` crates.

README.md
```diff
@@ -133,6 +133,78 @@ fn main() {
...
 ```
 
+## Example: Futures + TLS
+
+By default, tarpc uses a `TcpStream` for communication, however you can also opt to use a
+`TlsStream<TcpStream>` when the `tls` feature of `tarpc` is enabled.
+
+When using TLS, some additional information is required. You will need to make [`TlsAcceptor`] and
+`TlsClientContext` structs; `TlsClientContext` requires a [`TlsConnector`]. The [`TlsAcceptor`] and
+[`TlsConnector`] structs are defined in the [native-tls] crate which is exposed by `tarpc`.
```
compressed: Yes, that's the direction I was going. I've seen libraries that have public interfaces which require structs from other crates generally do this. You can see in [reqwest](https://github.com/seanmonstar/reqwest/blob/master/src/lib.rs#L100) and [hyper](https://github.com/hyperium/hyper/blob/master/src/lib.rs#L36). I think the best approach would be to just re-export only the external types we need, not the whole crate. I'll make that change when I do my updates.

compressed:
Thanks for all the comments. I'll work on these tomorrow!

src/server.rs
```diff
@@ -50,34 +75,48 @@ pub fn listen<S, Req, Resp, E>(new_service: S, addr: SocketAddr, options: Option
           Resp: Serialize + 'static,
           E: Serialize + 'static
 {
+    // similar to the client, since `Options` is not `Send`, we take the `TlsAcceptor` when it is
```
shaladdle: nit: Capitalize

src/server.rs
```diff
@@ -50,34 +75,48 @@ pub fn listen<S, Req, Resp, E>(new_service: S, addr: SocketAddr, options: Option
...
 /// Spawns a service that binds to the given address using the given handle.
 #[doc(hidden)]
-pub fn listen_with<S, Req, Resp, E>(new_service: S,
-                                    addr: SocketAddr,
-                                    handle: Handle)
-                                    -> io::Result<SocketAddr>
+fn listen_with<S, Req, Resp, E>(new_service: S,
+                                addr: SocketAddr,
+                                handle: Handle,
+                                _acceptor: Acceptor)
```
shaladdle: Why `_`? Is this unused with tls off or something?

src/client.rs
compressed: Yes, this is very good point. I think we can just have another function which takes a `TlsConnector`, that way people can build it however they want (e.g. overriding certificates, etc. and whatever else `native-tis` exposes in the future.

``` pub fn from_connector<S: Into<String>>(domain: S, tls_connector: TlsConnector) -> Self```

File: .travis.yml (hunk 14 above)
Comment (tikue) [inline .travis.yml]: What about coveralls? Looks like it needs to have the tls feature enabled.

File: src/client.rs (hunk 4 above)
Comment (tikue) [inline src/client.rs]: If you can, test this with clippy first. Not sure if it will complain or not. I hope not.

**Comment by tikue** [comment]
I can't figure out how to reply to one of @shaladdle's comments, so I'm just writing it here. It would be great if a non-tls client trying to connect to a tls server receives a specific tls-related error, like `Unauthenticated`. Is that possible?

File: src/lib.rs [hunk 20]
```diff
@@ -59,6 +60,53 @@
 //! }
 //! ```
 //!
+//! Example usage with TLS:
+//!
+#![cfg_attr(feature = "tls", doc = " ```rust,no_run")]
```
compressed: These provide some configurability for what runs during rustdoc's test.

So for these two:

```rust
#![cfg_attr(feature = "tls", doc = " ```ignore")]
#![cfg_attr(not(feature = "tls"), doc = " ```")]
```

It's interpreted something like this:

```
if feature != tls { return //! ``` };
if feature == tls { return //! ```ignore };
```

`ignore` just means don't run this test case at all. `no_run` is going to _compile_ the test, but not actually run it. We want this for the tls use case because we need to extra stuff like in the test case to add the certs to the cert chain, which we don't need to here in the rustdoc.

For the tcp side actually we can run that now even with tls enabled, so the first set of conditions can be removed.

src/server.rs
compressed: Yes, that's correct. We don't use `acceptor` in the tcp workflow.

src/macros.rs
compressed: We could just call it `unwrap!` I guess.

src/server.rs
compressed: Done

src/client.rs
```diff
@@ -227,6 +352,7 @@ pub mod future {
             ConnectFuture { inner: future::Either::B(rx.map_err(panic as fn(_) -> _).flatten()) }
         }
     }
+
```
compressed: Good catch. I have removed it.

src/client.rs
```diff
@@ -194,20 +288,51 @@ pub mod future {
...
                         .then(move |result| {
                             tx.complete(result);
                             Ok(())
                         })
                 }
             };
+
             let rx = match options.reactor {
                 Some(Reactor::Handle(handle)) => {
-                    let tcp = TcpStream::connect(&addr, &handle).map(MultiplexConnect::new(handle));
```
compressed: Good idea. I've implemented that.

README.md
```diff
@@ -133,6 +133,78 @@ fn main() {
 }
 ```
 
+## Example: Futures + TLS
+
+By default, tarpc uses a `TcpStream` for communication, however you can also opt to use a
+`TlsStream<TcpStream>` when the `tls` feature of `tarpc` is enabled.
```
compressed: I've added some additional context now. Let me know how this looks.

src/server.rs
compressed: Done

src/server.rs
compressed: Done

.travis.yml
compressed: Ideally yes... However, coveralls is broken already in `travis-cargo`. I think there are a few PRs to fix it in their repo. Maybe we should have a separate PR just to remove `travis-cargo`. It may be easier than using that tool.

compressed:
I've made updates to incorporate most of the comments (and also updated to latest master).

Here are the outstanding items/discussion points (I think!):
1. Test cases / error conditions: I added one test case before that tests both tcp and tls where they are configured properly, so that functionality is validated. However, I didn't add a test case for misconfigured clients. If you use a TLS client on a TCP server, the client connection just waits indefinitely, probably because the client is waiting for the tls handshake to begin. If you use a TCP client on a TLS server, the server logs an error when the client makes an RPC call and the client waits indefinitely (`ERROR:tarpc::server: While processing incoming connections: connection closed via error`, from here: https://github.com/google/tarpc/blob/master/src/server.rs#L98). Ideally, it would be best to catch these misconfigurations at compile time, but I'm not sure the best way to do that. I was thinking we could require some reference from the server to make the client, but even that won't work because the server struct itself could be used to make both a tcp and a tls server... Not sure here.
1. `TlsClientContext` name? I kept it as-is for now. Since we don't have `tls::client` etc., I thought we should keep client in the name. It also goes along with the other tls types: `TlsAcceptor`, `TlsConnector`, but open to change. 
1. Lots of `cfg`s. I'm not sure what to do about this either. The only way to eliminate most of them would be to make the interfaces generic over all stream types but we tried that before in the earlier versions of this PR, but then macro itself becomes quite unwieldly which I think is even worse.

tikue:
> TlsClientContext name? I kept it as-is for now. Since we don't have tls::client etc., I thought we should keep client in the name. It also goes along with the other tls types: TlsAcceptor, TlsConnector, but open to change.

I still think it should be named `Context`. It can be used as `client::tls::Context` or aliased `as ClientTlsContext` or `TlsClientContext`. Also, I don't think there are good reasons to reexport it from the crate root.

compressed:
Okay thanks! Works for me. I've made that change now too.

src/lib.rs
```diff
@@ -138,3 +188,16 @@ enum Reactor {
     Handle(reactor::Handle),
     Remote(reactor::Remote),
 }
+
+cfg_if! {
+    if #[cfg(feature = "tls")] {
+        extern crate tokio_tls;
+        extern crate native_tls;
+
+        /// Re-exported TLS-related types
```
shaladdle: Why do these need to be re-exported?

src/client.rs
```diff
@@ -19,8 +21,41 @@ use tokio_service::Service;
...
+    /// TLS context for client
+    pub struct Context {
+        /// Domain to connect to
+        pub domain: String,
+        /// TLS connector
+        pub tls_connector: TlsConnector,
+    }
+
+    impl Context {
+        /// Try to construct a new `Context`, providing the domain the client will
```
shaladdle: s/providing/provided?

src/lib.rs
compressed: Because we have public interfaces now that use all these types, except `Pkcs12`, but you need that type to make the `TlsAcceptor` itself.

src/client.rs
compressed: Good point, actually this comment is wrong. It's the domain used for validation, not connection. I will update that.

compressed:
Great! I've updated the comments for `domain` in the most recent commit to be more accurate.

tikue:
[Relevant issue](https://github.com/rust-lang/cargo/issues/2064)

src/lib.rs
```diff
@@ -138,3 +188,16 @@ enum Reactor {
...
+        pub mod tls {
+            pub use native_tls::Error as NativeTlsError;
```
tikue: what about renaming the module `native_tls` and then `pub use native_tls::Error`? (and then just add it to the below `pub use` statement)

src/macros.rs
```diff
@@ -696,21 +696,169 @@ mod syntax_test {
...
         rpc add(x: i32, y: i32) -> i32;
         rpc hey(name: String) -> String;
     }
 
+    cfg_if! {
+        if #[cfg(feature = "tls")] {
+            const DOMAIN: &'static str = "foobar.com";
+
+            use client::tls::Context;
+            use ::native_tls::{Pkcs12, TlsAcceptor, TlsConnector};
```
tikue: I think the `::` prefix is redundant?

compressed: Fixed.

src/lib.rs
compressed: Done. Used `extern crate native_tls as native_tls_inner` to avoid the name spacing problem.

compressed:
You're welcome!  I'm excited for it to land. Thanks for all the comments and working with me on it.

I just made one more minor change to add `Error` to the `use` above this line: https://github.com/compressed/tarpc/blob/c6bc6a871be294303942f7d70980f49bbdc811ab/src/client.rs#L46 so then the type can just be `Error` not `::native_tls::Error` in the `new` fn so that's why it's building again.

shaladdle:
+1, this is great, @compressed. Thanks so much for contributing :)

PR #82, opened 2017-01-23
Make sure synchronous RPCs are wrapped in a lazy future.

tikue:
Some future-returning fns implicitly require the presence of an execution task. Wrapping in a lazy future ensures that by the time the future is polled, there is a task present.

### DIFF
--- a/examples/throughput.rs
+++ b/examples/throughput.rs
@@ -30,7 +30,7 @@ lazy_static! {
 fn gen_vec(size: usize) -> Vec<u8> {
     let mut vec: Vec<u8> = Vec::with_capacity(size);
     for i in 0..size {
-        vec.push((i % 1 << 8) as u8);
+        vec.push(((i % 2) << 8) as u8);
     }
     vec
 }
@@ -98,7 +98,7 @@ fn bench_tcp(target: u64) {
 
 fn main() {
     let _ = env_logger::init();
-    &*BUF; // to non-lazily initialize it.
+    let _ = *BUF; // to non-lazily initialize it.
     bench_tcp(256 << 20);
     bench_tarpc(256 << 20);
 }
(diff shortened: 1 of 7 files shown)


examples/throughput.rs
```diff
@@ -99,7 +98,7 @@ fn bench_tcp(target: u64) {
 
 fn main() {
     let _ = env_logger::init();
-    &*BUF; // to non-lazily initialize it.
+    let _ = *BUF; // to non-lazily initialize it.
```
shaladdle: nit: Capitalize

src/client.rs
```diff
@@ -112,14 +112,14 @@ impl Options {
 
 /// Exposes a trait for connecting asynchronously to servers.
 pub mod future {
+    use super::{Client, Options};
```
shaladdle: rustfmt changing its mind since last version again?

src/client.rs
```diff
@@ -140,6 +140,7 @@ pub mod future {
               Resp: Deserialize + 'static,
               E: Deserialize + 'static,
     {
+        #[allow(unknown_lints, type_complexity)]
```
shaladdle: What's this do?

src/macros.rs
```diff
@@ -530,16 +534,17 @@ macro_rules! service {
         pub struct SyncClient(FutureClient);
 
         impl $crate::client::sync::Connect for SyncClient {
-            fn connect<A>(addr: A, options: $crate::client::Options)
+            fn connect<A>(_addr: A, _opts: $crate::client::Options)
```
shaladdle: Why the underscores? To avoid naming collisions? This will remove unused variable checks I think.

src/client.rs
tikue: Guess so?

src/client.rs
tikue: Tells rustc to shut up about lints it doesn't know about, and tells clippy to shut up about its type complexity lint.

examples/throughput.rs
tikue: Done.

src/macros.rs
tikue: Oh, I didn't think about that. I guess an underscore suffix would accomplish the same thing.

PR #91, opened 2017-02-02
Temporary workaround for compiler bugs in impl Trait feature.

tikue:
We will want to revert this as soon as possible. There's not really a good way to test whether it's fixed because we don't even have a rustc bug right now to track.

cc @jonhoo

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -268,9 +268,10 @@ macro_rules! service {
         )*
     ) => {
 
+        #[doc(hidden)]
         #[allow(non_camel_case_types, unused)]
         #[derive(Debug)]
-        enum tarpc_service_Request__ {
+        pub enum tarpc_service_Request__ {
             NotIrrefutable(()),
             $(
                 $fn_name(( $($in_,)* ))
@@ -280,9 +281,10 @@ macro_rules! service {
         impl_deserialize!(tarpc_service_Request__, NotIrrefutable(()) $($fn_name(($($in_),*)))*);
         impl_serialize!(tarpc_service_Request__, {}, NotIrrefutable(()) $($fn_name(($($in_),*)))*);
 
+        #[doc(hidden)]
         #[allow(non_camel_case_types, unused)]
         #[derive(Debug)]
-        enum tarpc_service_Response__ {
+        pub enum tarpc_service_Response__ {
             NotIrrefutable(()),
             $(
                 $fn_name($out)
@@ -292,9 +294,10 @@ macro_rules! service {
         impl_deserialize!(tarpc_service_Response__, NotIrrefutable(()) $($fn_name($out))*);
         impl_serialize!(tarpc_service_Response__, {}, NotIrrefutable(()) $($fn_name($out))*);
 
+        #[doc(hidden)]
         #[allow(non_camel_case_types, unused)]
         #[derive(Debug)]
-        enum tarpc_service_Error__ {
+        pub enum tarpc_service_Error__ {
             NotIrrefutable(()),
             $(
                 $fn_name($error)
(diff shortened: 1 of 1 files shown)


mcmathja:
Was this about a linker error when trying to implement FutureClient from a separate crate? Was running into issues with that the other night. Tracked this down, if it's of any assistance: https://github.com/rust-lang/rust/issues/35870. Modifying the macro to mark service methods inline resolved the issue in my case.

compressed:
That's interesting. That issue also talks about some visibility issues. I wonder if that is actually the underlying problem since the full type has some types that needed to be public with this change.

Maybe just making those public would make it go away.

jonhoo:
@mcmathja yeah, that bug came up in our discussion on gitter yesterday too — thanks for confirming!

@compressed I just tried locally reverting the signature change in 9d552e48a482db8c130cdbacfb8e0252bea7fdcb, but keeping the visibility modifiers, and that then still causes the linking error for me unfortunately :/

tikue:
Thanks for the info @mcmathja! Do you happen to have a reproduction you can share? I'd love to open a rustc issue.

mcmathja:
Played around with this for a bit and the issue is more subtle than I thought. The calls to `unreachable!()` [here](https://github.com/google/tarpc/blob/master/src/macros.rs#L656) and [here](https://github.com/google/tarpc/blob/master/src/macros.rs#L667) are getting expanded to the standard panic call, which runs something like:

```rust
::rt::begin_panic("internal error: entered unreachable code",
{
    static _FILE_LINE: (&'static str, u32) = ("src/lib.rs", 7u32);
     &_FILE_LINE
})
```

For me at least, it seems to be those `_FILE_LINE` references that are creating difficulties:

```
  = note: Undefined symbols for architecture x86_64:
            "service::FutureClient::test::_$u7b$$u7b$closure$u7d$$u7d$::_FILE_LINE::hd619a832b2c2aa1b", referenced from:
                service::FutureClient::test::_$u7b$$u7b$closure$u7d$$u7d$::h4a4885791567c19d in client-b336e68e002de9fe.0.o
            "service::FutureClient::test::_$u7b$$u7b$closure$u7d$$u7d$::_FILE_LINE::hcb35c9d631e99707", referenced from:
                service::FutureClient::test::_$u7b$$u7b$closure$u7d$$u7d$::h4a4885791567c19d in client-b336e68e002de9fe.0.o
```

Replacing the macros with concrete errors resolves the issue. It's beyond me why only those two calls need to be replaced, I haven't been able to set up a more generic reproduction. (Have the one I used to generate those errors [here](https://github.com/mcmathja/tarpc_repro)). Looks like a different bug from the one I linked earlier in any case.

jonhoo:
Yup, that looks like [my error](http://pastebin.com/gLvs4DDc) too.

jonhoo:
Interestingly, I get a *bunch* of these errors if I try to compile [`alacritty`](https://github.com/jwilm/alacritty) on `beta` (after a `cargo update`; haven't tried without).

jonhoo:
So here's something interesting: I now have https://github.com/mit-pdos/distributary/commit/b7abc6ed93e1abc16f3054df8fc292a4d7683e96 compiling on one machine, and failing with a linker error on another, just using `impl Trait` in *my* crate, and on the same nightly (2017-03-03). Digging into it now. Will report back what I find.

jonhoo:
Travis also complains: https://travis-ci.com/mit-pdos/distributary/builds/41405053

jonhoo:
A `cargo update` seems to have inexplicably fixed the problem, even though no dependencies changes. `cargo build --release` still fails though. And on more machines..

jonhoo:
Following this up in https://github.com/rust-lang/rust/issues/35870#issuecomment-284948755

PR #92, opened 2017-02-02
Change SyncClient to drive its own execution with an internal reactor::Core.

### LINKED ISSUE #88: Document that with default options, all clients share single event loop
`Options::default()` sets `Options.reactor = None`, which causes `Client::connect` to use [`REMOTE.spawn`](https://github.com/google/tarpc/blob/master/src/client.rs#L345). To my great surprise, `REMOTE` is a [static](https://github.com/google/tarpc/blob/master/src/lib.rs#L168) event loop driven by [a single thread](https://github.com/google/tarpc/blob/master/src/lib.rs#L176). Thus, if `Options::default()` is used, all RPC calls block on a single thread. This should be clearly documented, as that thread will quickly become a big bottleneck for high performance systems.

### LINKED ISSUE #89: SyncClient with event loop driven by current thread
Following the observations in #88, I wanted to modify the code such that the thread *making* an RPC call is the one that drives the event loop. However, there doesn't seem to any way to do this with `SyncClient` as far as I can tell? When you make the function call representing an RPC method, you have no way to *also* make that thread drive the core until the call completes? Of course, with a `FutureClient` you could implement this yourself pretty trivially, but having this be a supported mode of operation for `SyncClient` would be really nice.

tikue:
Adds a `reactor::Core` to `SyncClient` to enable the client to drive its own execution, as one would expect.

Previously, the reactor had to be driven on a separate thread, which was confusing.
This has a couple notable side effects:
  1. `SyncClient` is no longer `Clone`.  This is because `reactor::Core`
     is not `Clone`, and creating one is not infallible
     (`Core::new` returns a `Result`). `try_clone` is added in #96 
  2. `SyncClient` does not use the user-specified `client::Options::handle` or
     `client::Options::remote`, because it constructs its own reactor. This problem will be exacerbated if we do end up making the `Handle` arg of `Options` required for future clients.
  3. `SyncClient` rpcs now take `&mut self` because `reactor::Core::run` takes `&mut self`. I don't see any issue here.

fixes #88, #89 

There was also a bug in tls connecting when using a `Handle` (instead of a `Remote` which is used in all current tests). I refactored the client to use the same code in both code paths, and at the same time was able to remove a lot of unneeded code.

### DIFF
--- a/benches/latency.rs
+++ b/benches/latency.rs
@@ -38,8 +38,13 @@ impl FutureService for Server {
 #[bench]
 fn latency(bencher: &mut Bencher) {
     let _ = env_logger::init();
-    let addr = Server.listen("localhost:0".first_socket_addr(), server::Options::default()).wait().unwrap();
+    let addr = Server.listen("localhost:0".first_socket_addr(),
+                server::Options::default())
+        .wait()
+        .unwrap();
     let client = SyncClient::connect(addr, client::Options::default()).unwrap();
 
-    bencher.iter(|| { client.ack().unwrap(); });
+    bencher.iter(|| {
+        client.ack().unwrap();
+    });
 }
--- a/examples/pubsub.rs
+++ b/examples/pubsub.rs
@@ -122,7 +122,7 @@ fn main() {
         .wait()
         .unwrap();
 
-    let publisher_client =
+    let mut publisher_client =
         publisher::SyncClient::connect(publisher_addr, client::Options::default()).unwrap();
 
     let subscriber1 = Subscriber::listen(0);
--- a/examples/readme_errors.rs
+++ b/examples/readme_errors.rs
@@ -51,7 +51,7 @@ impl SyncService for HelloServer {
 
 fn main() {
     let addr = HelloServer.listen("localhost:10000", server::Options::default()).unwrap();
-    let client = SyncClient::connect(addr, client::Options::default()).unwrap();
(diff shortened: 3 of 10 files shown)


tikue:
Regarding 1, this is a pretty significant change, but I think it's reasonable because:

1. `SyncClient` is easy mode, so if you need more advanced features you can always drop down to `FutureClient`.
2. A user can always connect multiple times.
3. Not being able to drive execution on the same thread that's blocking on the rpc is a bigger flaw that not having `Clone`.

tikue:
Regarding 2, we could have `sync::Options` and `future::Options`, but it might be better ergonomically to just document the fact that the handle option is not used by the sync client. I don't have a strong opinion here.

jonhoo:
Re `Clone`, I think a neat middle-ground would be to add a `try_clone` method to `SyncClient` which returns a `Result<Self>`. It's not as ergonomic as having the standard `Clone` trait implemented, but makes the removal of `Clone` somewhat less annoying.

Re `Options`, I agree that documenting that it is ignored is probably a decent first step.

tikue:
@jonhoo yes that's what I'm thinking as well. I'm thinking of renaming the Connect traits to `ClientExt` so that future extensions like this can be made more easily.

shaladdle:
Sorry for the delay. `try_clone` sounds like a good solution to the problem. There definitely needs to be some way to use the same underlying TcpConnection from multiple threads in sync world.

tikue:
Merged master.

Issue #93, opened 2017-02-02
Move to serde 0.9


tikue:
Bincode supports it now so there shouldn't be any blockers.

cc @compressed

tikue:
Fixed by #94

PR #94, opened 2017-02-02
Upgrade serde to 0.9

compressed:
I've made changes to support serde 0.9. Serde 0.9 is not compatible with serde 0.8 so this is a breaking change.

I also upgraded byteorder to 1.0 (no changes were needed for that).


tikue:
Fixes #93

PR #96, opened 2017-02-03
Remove the `Send` bound from `FutureService`

### LINKED ISSUE #95: FutureService should not require `Send`
Only `SyncService` needs to be `Send`.

tikue:
**This builds on #92 (so it should be reviewed after it).**

A few changes are included in this PR. The first listed is the main change and the others are necessary consequences of the first:

1. Removes the `Send` bound from `FutureService`. The `Send` bound is still required for `SyncService`, since clones are sent to new threads for each request. (This is more fodder for the argument that there should be a distinct `Options` struct for each combination of async/sync and client/server.)
2. Makes specifying a reactor handle mandatory for server options; it no longer defaults to the tarpc global `REMOTE` because that would require services to be `Send`.
2. Adds the ability to specify a `reactor::Core` (instead of `Remote` or `Handle`) on `client::Options`. This is important because with just #92 it's not possible to drive other futures with a `SyncClient`. (Not sure how frequently that's needed in *real* code, but combined with change 1., it's hard to write examples without it.).
3. Makes `FutureService::listen` return an `io::Result`rather than a `Future`; the future was never actually necessary and had the unintended consequence of making `SyncService::listen` deadlock when the options specified a handle (because that means the reactor driving the service lives on the same thread that `SyncService` is waiting on).

Additionally, this PR renames the client connect traits to `ClientExt` and adds the missing `try_clone` to the renamed `client::sync::ClientExt`.

Fixes #95

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ extern crate tokio_core;
 
 use futures::Future;
 use tarpc::{client, server};
-use tarpc::client::future::Connect;
+use tarpc::client::future::ClientExt;
 use tarpc::util::{FirstSocketAddr, Never};
 use tokio_core::reactor;
 
@@ -174,7 +174,7 @@ extern crate tokio_core;
 
 use futures::Future;
 use tarpc::{client, server};
-use tarpc::client::future::Connect;
+use tarpc::client::future::ClientExt;
 use tarpc::util::{FirstSocketAddr, Never};
 use tokio_core::reactor;
 use tarpc::native_tls::{Pkcs12, TlsAcceptor};
--- a/README.md
+++ b/README.md
@@ -50,10 +50,12 @@ tarpc-plugins = { git = "https://github.com/google/tarpc" }
 extern crate futures;
 #[macro_use]
 extern crate tarpc;
+extern crate tokio_core;
 
 use tarpc::{client, server};
 use tarpc::client::sync::Connect;
-use tarpc::util::Never;
+use tarpc::util::{FirstSocketAddr, Never};
+use tokio_core::reactor;
 
 service! {
     rpc hello(name: String) -> String;
@@ -69,9 +71,11 @@ impl SyncService for HelloServer {
 }
 
 fn main() {
-    let addr = "localhost:10000";
-    HelloServer.listen(addr, server::Options::default()).unwrap();
-    let client = SyncClient::connect(addr, client::Options::default()).unwrap();
+    let reactor = reactor::Core::new().unwrap();
+    let addr = HelloServer.listen("localhost:0".first_socket_addr(),
+                                  server::Options::from(reactor.handle()))
+                          .unwrap();
+    let mut client = SyncClient::connect(addr, client::Options::default().core(reactor)).unwrap();
     println!("{}", client.hello("Mom".to_string()).unwrap());
 }
 ```
@@ -121,9 +125,10 @@ impl FutureService for HelloServer {
 }
 
 fn main() {
-    let addr = "localhost:10000".first_socket_addr();
     let mut core = reactor::Core::new().unwrap();
-    HelloServer.listen(addr, server::Options::default().handle(core.handle())).wait().unwrap();
+    let addr = HelloServer.listen("localhost:10000".first_socket_addr(),
+                                  server::Options::from(core.handle()))
+                          .unwrap();
     let options = client::Options::default().handle(core.handle());
     core.run(FutureClient::connect(addr, options)
             .map_err(tarpc::Error::from)
@@ -196,14 +201,16 @@ fn get_acceptor() -> TlsAcceptor {
 }
 
 fn main() {
-    let addr = "localhost:10000".first_socket_addr();
     let mut core = reactor::Core::new().unwrap();
     let acceptor = get_acceptor();
-    HelloServer.listen(addr, server::Options::default()
-                                 .handle(core.handle())
-                                 .tls(acceptor)).wait().unwrap();
-    let options = client::Options::default().handle(core.handle()
-                      .tls(client::tls::Context::new("foobar.com").unwrap()));
+    let addr = HelloServer.listen("localhost:10000".first_socket_addr(),
+                                  server::Options::default()
+                                      .handle(core.handle())
+                                      .tls(acceptor))
+                          .unwrap();
+    let options = client::Options::default()
+                                   .handle(core.handle())
+                                   .tls(client::tls::Context::new("foobar.com").unwrap()));
     core.run(FutureClient::connect(addr, options)
             .map_err(tarpc::Error::from)
             .and_then(|client| client.hello("Mom".to_string()))
--- a/README.md
+++ b/README.md
@@ -204,8 +204,7 @@ fn main() {
     let mut core = reactor::Core::new().unwrap();
     let acceptor = get_acceptor();
     let addr = HelloServer.listen("localhost:10000".first_socket_addr(),
-                                  server::Options::default()
-                                      .handle(core.handle())
+                                  server::Options::from(core.handle())
                                       .tls(acceptor))
                           .unwrap();
     let options = client::Options::default()
--- a/README.md
+++ b/README.md
@@ -73,9 +73,9 @@ impl SyncService for HelloServer {
 fn main() {
     let reactor = reactor::Core::new().unwrap();
     let addr = HelloServer.listen("localhost:0".first_socket_addr(),
-                                  server::Options::from(reactor.handle()))
+                                  server::Options::default())
                           .unwrap();
-    let mut client = SyncClient::connect(addr, client::Options::default().core(reactor)).unwrap();
+    let mut client = SyncClient::connect(addr, client::Options::default()).unwrap();
     println!("{}", client.hello("Mom".to_string()).unwrap());
 }
 ```
--- a/benches/latency.rs
+++ b/benches/latency.rs
@@ -41,11 +41,9 @@ fn latency(bencher: &mut Bencher) {
     let _ = env_logger::init();
     let reactor = reactor::Core::new().unwrap();
     let addr = Server.listen("localhost:0".first_socket_addr(),
-                             server::Options::from(reactor.handle()))
-                     .unwrap();
-    let mut client = SyncClient::connect(addr, client::Options::default().core(reactor)).unwrap();
+                server::Options::from(reactor.handle()))
+        .unwrap();
+    let mut client = SyncClient::connect(addr, client::Options::default()).unwrap();
 
-    bencher.iter(|| {
-        client.ack().unwrap();
-    });
+    bencher.iter(|| { client.ack().unwrap(); });
 }
--- a/benches/latency.rs
+++ b/benches/latency.rs
@@ -12,13 +12,14 @@ extern crate tarpc;
 extern crate test;
 extern crate env_logger;
 extern crate futures;
+extern crate tokio_core;
 
-use futures::Future;
 use tarpc::{client, server};
 use tarpc::client::sync::Connect;
 use tarpc::util::{FirstSocketAddr, Never};
 #[cfg(test)]
 use test::Bencher;
+use tokio_core::reactor;
 
 service! {
     rpc ack();
@@ -38,11 +39,11 @@ impl FutureService for Server {
 #[bench]
 fn latency(bencher: &mut Bencher) {
     let _ = env_logger::init();
+    let reactor = reactor::Core::new().unwrap();
     let addr = Server.listen("localhost:0".first_socket_addr(),
-                server::Options::default())
-        .wait()
-        .unwrap();
-    let client = SyncClient::connect(addr, client::Options::default()).unwrap();
+                             server::Options::from(reactor.handle()))
+                     .unwrap();
+    let mut client = SyncClient::connect(addr, client::Options::default().core(reactor)).unwrap();
 
     bencher.iter(|| {
         client.ack().unwrap();
--- a/benches/latency.rs
+++ b/benches/latency.rs
@@ -19,7 +19,6 @@ use tarpc::client::sync::ClientExt;
 use tarpc::util::{FirstSocketAddr, Never};
 #[cfg(test)]
 use test::Bencher;
-use tokio_core::reactor;
 
 service! {
     rpc ack();
@@ -28,10 +27,9 @@ service! {
 #[derive(Clone)]
 struct Server;
 
-impl FutureService for Server {
-    type AckFut = futures::Finished<(), Never>;
-    fn ack(&self) -> Self::AckFut {
-        futures::finished(())
+impl SyncService for Server {
+    fn ack(&self) -> Result<(), Never> {
+        Ok(())
     }
 }
 
@@ -39,11 +37,10 @@ impl FutureService for Server {
 #[bench]
 fn latency(bencher: &mut Bencher) {
     let _ = env_logger::init();
-    let reactor = reactor::Core::new().unwrap();
     let addr = Server.listen("localhost:0".first_socket_addr(),
-                server::Options::from(reactor.handle()))
+                             server::Options::default())
         .unwrap();
     let mut client = SyncClient::connect(addr, client::Options::default()).unwrap();
 
-    bencher.iter(|| { client.ack().unwrap(); });
+    bencher.iter(|| client.ack().unwrap());
 }
--- a/benches/latency.rs
+++ b/benches/latency.rs
@@ -15,7 +15,7 @@ extern crate futures;
 extern crate tokio_core;
 
 use tarpc::{client, server};
-use tarpc::client::sync::Connect;
+use tarpc::client::sync::ClientExt;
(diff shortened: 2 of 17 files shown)


README.md
```diff
@@ -196,14 +201,16 @@ fn get_acceptor() -> TlsAcceptor {
...
-    let addr = "localhost:10000".first_socket_addr();
     let mut core = reactor::Core::new().unwrap();
     let acceptor = get_acceptor();
-    HelloServer.listen(addr, server::Options::default()
-                                 .handle(core.handle())
-                                 .tls(acceptor)).wait().unwrap();
-    let options = client::Options::default().handle(core.handle()
-                      .tls(client::tls::Context::new("foobar.com").unwrap()));
+    let addr = HelloServer.listen("localhost:10000".first_socket_addr(),
+                                  server::Options::default()
```
compressed: I think this should be `server::Options::from(core.handle())...` like the other ones since `Default` is gone.

tikue: Fixed, thanks!

tikue:
Fixed merge conflicts.

examples/pubsub.rs
```diff
@@ -57,39 +59,39 @@ impl subscriber::FutureService for Subscriber {
...
-            .wait()
+            .listen("localhost:0".first_socket_addr(), options)
             .unwrap()
     }
 }
 
 #[derive(Clone, Debug)]
 struct Publisher {
-    clients: Arc<Mutex<HashMap<u32, subscriber::FutureClient>>>,
+    clients: Rc<RefCell<HashMap<u32, subscriber::FutureClient>>>,
```
tikue: Not to resist adding a comment, but I think it's worth pointing out that in Rust thread safety is guaranteed by the compiler, so using `Rc<RefCell<>>` over `Arc<Mutex<>>` is always the right thing as long as the compiler allows it. In this case, `Rc` is not `Send`, which is why I had to remove the `Send` bound before I could replace `Arc`.

tikue:
Regarding 1

> I'm not sure what the cons of this approach are
The obvious con is that you'd need significantly more boilerplate for example code. You'd have to spin up a thread to drive the services because it wouldn't be possible to drive the server on the same thread as the sync client if it doesn't share a reactor. You'd also have to consider how to actually drive the reactor, which typically means running a future that never resolves or something like `loop { reactor.run_once(); }`.

Regarding 2
> People should only use this feature At Their Own Risk™
Can you elaborate on what the risks are? You're worried that they'll use different reactors to drive the service and client, but by removing this capability that only makes that problem more pronounced, no?

examples/two_clients.rs
```diff
@@ -59,17 +60,25 @@ macro_rules! pos {
...
-        .unwrap();
-
-    let bar_client = bar::SyncClient::connect(bar_addr, client::Options::default()).unwrap();
-    let baz_client = baz::SyncClient::connect(baz_addr, client::Options::default()).unwrap();
+    let mut bar_client = {
+        let reactor = reactor::Core::new().unwrap();
+        let addr = Bar.listen("localhost:0".first_socket_addr(),
+                              server::Options::from(reactor.handle()))
+                      .unwrap();
+        // TODO: Need to set up each client with its own reactor. Should it be shareable across
```
tikue: The "need to set up" was meant to describe the status quo, not what needs to be changed.

src/server.rs
```diff
@@ -33,23 +32,26 @@ enum Acceptor {
 }
 
 /// Additional options to configure how the server operates.
-#[derive(Default)]
 pub struct Options {
-    reactor: Option<Reactor>,
+    handle: reactor::Handle,
```
tikue: Seems reasonable.

src/server.rs
```diff
@@ -84,47 +86,23 @@ pub fn listen<S, Req, Resp, E>(new_service: S, addr: SocketAddr, options: Option
...
                         Response = Response<Resp, E>,
-                        Error = io::Error> + Send + 'static,
+                        Error = io::Error> + 'static,
           Req: Deserialize + 'static,
           Resp: Serialize + 'static,
           E: Serialize + 'static
 {
     let listener = listener(&addr, &handle)?;
-    let addr = listener.local_addr()?;
+    let addr = listener.local_addr();
```
tikue: Good question. I don't know the failure modes of `local_addr` but it might be independent of the rest of server initialization. e.g. maybe a `local_addr` doesn't actually exist for some reason?

README.md
```diff
@@ -69,9 +71,11 @@ impl SyncService for HelloServer {
 }
 
 fn main() {
-    let addr = "localhost:10000";
-    HelloServer.listen(addr, server::Options::default()).unwrap();
-    let client = SyncClient::connect(addr, client::Options::default()).unwrap();
+    let reactor = reactor::Core::new().unwrap();
```
tikue: I prefer reactor but I don't know what's more standard.

README.md
```diff
@@ -69,9 +71,11 @@ impl SyncService for HelloServer {
...
+    let addr = HelloServer.listen("localhost:0".first_socket_addr(),
+                                  server::Options::from(reactor.handle()))
+                          .unwrap();
+    let mut client = SyncClient::connect(addr, client::Options::default().core(reactor)).unwrap();
```
tikue: This is the easiest way to get up and running in as little code as possible. There's precedence in crates for taking the easy route in examples, for example playing fast and loose with `unwrap`. There's always a balance between "get working quickly" and "the right way", and in this particular case it's not even wrong per se, just quirky.

shaladdle:
> As a general point, using tarpc at this point requires knowledge of tokio regardless of whether you use the sync or future client,

Here's the disconnect I guess. I think it makes more sense if users didn't need to learn about tokio or how to use it in order to figure out how to use the tarpc sync API. The point of the sync API is to allow users to write blocking code to call and expose tarpc services. I'd like the sync API to be designed with that as the driving goal, regardless of what the implementation is underneath.

If this means that we wrap the `FutureService` in some `Server` type which exposes a `run()` method, that sounds fine to me. We could also have another abstraction which holds the reactor and allows users to register multiple services with it. So you'd write code like this in your actual server:

```rust
fn main() {
  let server = tarpc::Server::new();
  server.bind(HelloService::new(), "localhost:1");
  server.bind(GoodByeService::new(), "localhost:2");

  // Blocks until something catastrophic happens. Underneath this runs the reactor.
  server.run().unwrap();
}
```

This is just brainstorming. This PR may not be the right place to come to a conclusion about this stuff. On that note, if you want to sidestep this issue, we could do something like remove the existing sync API until we've (mostly I, perhaps) had a chance to flesh out the details. I think these concerns are worth fleshing out, but maybe it's better if I do it as a side thing rather than blocking every PR you send because it affects the sync API and we don't have a design that we've agreed on guiding our way. What do you think?

> Can you elaborate on what the risks are? You're worried that they'll use different reactors to drive the server and client, but by removing this capability that only makes that problem more pronounced, no (they'd have no choice but to use different reactors to drive the server and client!)?

It's fine to have the ability to use different reactors to drive servers and clients. My objection stems from the fact that I don't see a real use case for having the client _drive_ the execution of a reactor with non-client futures running on it. I'd prefer to not include features that aren't driven by real use cases.

Beyond that, I can try to elaborate on risks, but without an example of how it's even supposed to be used, the things I come up with are crazy:

```rust
let reactor = Core::new();
// Maybe used for health checking
let ping_service = PingService::new();
ping_service.listen("localhost:1", reactor.handle());
let hello_service = HelloService::new();
hello_service.listen("localhost:2", reactor.handle());
client = tarpc::sync::Client::new("localhost:2", reactor);
loop {
  thread::sleep(Duration::from_secs(10));
  println!("{}", client.hello());
}
```

I feel silly just writing that down. Why do we even have that lever? Maybe I'm stupid for writing that code, but it seems like only a step further from the example code, which tends to be how people consume a new library. If someone new to tarpc and tokio comes along, they might wonder why `ping_service` only responds sometimes.

A separate meta point is that when I look at the examples as they are, I think they're not very practical. I think we should have examples that are short and sweet, but they should also be a reasonable starting point for someone trying to actually use tarpc. Most people trying to actually use tarpc would have a client and server binary, and have the two communicate. The server binary is going to have to block the main thread forever while the server handles connections, and the examples don't show how to do that. I noticed this while looking at the 'client runs the server's reactor core' code, and sort of took a step back and realized the disconnect here. This could also be addressed as another PR.

examples/pubsub.rs
shaladdle: Sure. It's fine to leave it with no comment. My thought process is like this: `RefCell` only performs runtime checks on mutable borrows, and therefore afaiu you can't quite declare victory once this compiles.

README.md
shaladdle: Either is fine with me, I think they should just be consistent across the examples.

src/server.rs
shaladdle: Can we err on the side of failing fast here and keep the `try` operator? (is that what it's called?)

tikue:
> Here's the disconnect I guess. I think it makes more sense if users didn't need to learn about tokio or how to use it in order to figure out how to use the tarpc sync API.

That's not possible unless you don't want to allow users to run multiple sync servers on a single thread.

tikue:
A special sync server construct that provides a thin abstraction on top of the reactor seems pointless to me, personally. It's literally a reactor by a different name?

tikue:
> This is just brainstorming. This PR may not be the right place to come to a conclusion about this stuff. On that note, if you want to sidestep this issue, we could do something like remove the existing sync API until we've (mostly I, perhaps) had a chance to flesh out the details. 

This leaves me conflicted. On the one hand, I think you're probably more motivated than me to find a good sync API, so removing it until the story is agreed upon is reasonable in that regard. On the other hand, people are using the sync API now and I don't think it's fair to leave them hanging indefinitely.

tikue:
> The server binary is going to have to block the main thread forever while the server handles connections, and the examples don't show how to do that.

This is a concern specific to sync APIs, because in tokio world you just run the reactor in a loop.

jonhoo:
> That's not possible unless you don't want to allow users to run multiple sync servers on a single thread.

So, I actually think this is exactly what `Sync` should do. The server should have its own internal reactor that is never exposed to the user (i.e., it is cooked up by `::new()`, not passed in), and `listen` should block indefinitely waiting for new requests. Similarly, `SyncClient` should have a separate internal reactor, which again is never exposed externally (it is created by `::new()`). Every call to `SyncClient` will block (as it is supposed to), and will internally drive that client's reactor until completion.

This provides a very straightforward API for simple use-cases, is easy to understand, and is decently performant for common-case uses (each server and client is run entirely by its own thread). For users that need more flexibility, such as wanting to multiplex services onto a single thread, or wanting server and client to be driven by the same thread, they should be directed to `FutureService` and `FutureClient`.

examples/pubsub.rs
jonhoo: Arguably this could actually be an `UnsafeCell`, which would make this explicit in the form of an `unsafe` where we document exactly why we believe there can be no other writers.

tikue:
I mostly agree, except `listen` can't block indefinitely because it needs to return the `SocketAddr`. I think it's fine if `listen` spawns a new thread that the server runs on.

jonhoo:
Either that, or have `listen` return a tuple of `(SocketAddr, ServerRunner)`, where `ServerRunner` is `#[must_use]`, and has only a single method, `run`, that consumes `self` and never returns. Obviously `ServerRunner` is not a great name, but you get the idea.

shaladdle:
> A special sync server construct that provides a thin abstraction on top of the reactor seems pointless to me, personally. It's literally a reactor by a different name?

Sure, there's some validity to that claim, but it's much simpler than a reactor from an API perspective. There's no context about tokio that needs to be known. It just has a few methods which have well defined behavior as laid out by the API. There 

> So, I actually think this is exactly what Sync should do. The server should have its own internal reactor that is never exposed to the user (i.e., it is cooked up by ::new(), not passed in), and listen should block indefinitely waiting for new requests. Similarly, SyncClient should have a separate internal reactor, which again is never exposed externally (it is created by ::new()). Every call to SyncClient will block (as it is supposed to), and will internally drive that client's reactor until completion.
>
> This provides a very straightforward API for simple use-cases, is easy to understand, and is decently performant for common-case uses (each server and client is run entirely by its own thread). For users that need more flexibility, such as wanting to multiplex services onto a single thread, or wanting server and client to be driven by the same thread, they should be directed to FutureService and FutureClient.

Yeah, I agree that this is a good place for the sync API to be for now. If people really want to stay on sync and fix some performance issue, we can address it then.

> Either that, or have listen return a tuple of (SocketAddr, ServerRunner), where ServerRunner is #[must_use], and has only a single method, run, that consumes self and never returns. Obviously ServerRunner is not a great name, but you get the idea.

This sounds similar to the `Server` idea I mentioned above. `server.bind(service, addr)` could return the bound socket addr.

src/server.rs
tikue: DOne.

src/server.rs
tikue: Done.

README.md
tikue: Done.

examples/pubsub.rs
tikue: I think `UnsafeCell` is pretty risky in concurrent code, even when that code is single threaded.

src/client.rs
```diff
@@ -300,90 +210,141 @@ pub mod future {
...
+        }
+    }
+
+    /// Extension methods for Clients.
+    pub trait ClientExt: Sized {
         /// Connects to a server located at the given address.
-        fn connect<A>(addr: A, options: Options) -> Result<Self, io::Error> where A: ToSocketAddrs;
+        fn connect<A>(addr: A, options: Options) -> io::Result<Self> where A: ToSocketAddrs;
+        /// Tries to clone the client.
+        fn try_clone(&self) -> io::Result<Self>;
```
tikue: Done. Annnnnd it doesn't work, because the cloned client's internals still have a handle to the original reactor. So...it seems there's no way to clone a client onto a different reactor.

benches/latency.rs
```diff
@@ -38,13 +39,11 @@ impl FutureService for Server {
 #[bench]
 fn latency(bencher: &mut Bencher) {
     let _ = env_logger::init();
+    let reactor = reactor::Core::new().unwrap();
     let addr = Server.listen("localhost:0".first_socket_addr(),
-                server::Options::default())
-        .wait()
+                server::Options::from(reactor.handle()))
```
shaladdle: I think from is gone now. Can you fix this one so it builds? I've filed #105 to fix travis.

benches/latency.rs
```diff
@@ -41,11 +41,9 @@ fn latency(bencher: &mut Bencher) {
     let _ = env_logger::init();
     let reactor = reactor::Core::new().unwrap();
     let addr = Server.listen("localhost:0".first_socket_addr(),
-                             server::Options::from(reactor.handle()))
-                     .unwrap();
-    let mut client = SyncClient::connect(addr, client::Options::default().core(reactor)).unwrap();
+                server::Options::from(reactor.handle()))
```
shaladdle: This is stale..

benches/latency.rs
shaladdle: You beat me to it.

examples/pubsub.rs
jonhoo: I mean, using `RefCell` will `panic!` instead, which I *guess* is a better outcome?

README.md
jonhoo: Correct me if I'm wrong, but isn't `reactor` no longer used in this function?

jonhoo:
I would personally prefer it if `listen` didn't spawn a new thread, but instead (as @shaladdle also proposed) give the caller a way to first `bind` (and not block), and then `listen` (and block forever).

tikue:
I think that's the agreed on plan. It was decided it didn't need to block this PR, as the design still needs to be fleshed out and the current solution should suffice in the interim.

tikue:
The main thing I want to enable via a sync server builder is running multiple heterogeneous servers on one thread. But making that work seems non-trivial due to coherence issues: they'd need to impl a common Bind trait, but the macro can't impl the trait for all T: SyncService because neither the trait nor the impl type is defined in the macro-calling crate.

jonhoo:
Hmm.. Do you think that's actually a common use-case? I feel like it would be perfectly fine to require users to use `FutureService` for that.

tikue:
If the only thing anyone else cares about is a simple way to block the main thread (why not just use `loop{}`?), that can be accomplished more easily.

tikue:
> Do you think that's actually a common use-case?

I think it depends a lot on who's using tarpc. In my experience,  rpc frameworks are commonly used in situations where servers are exposing many services. And if that's the case, it's best not to expend resources more than necessary. As you've said though, they can always just use futures, so maybe it's not worth spending too much time on it.

jonhoo:
I would say that *by far* the most common use-case for RPC is to expose only a single service. You want to split something up between a server and a bunch of clients. And I think that's the use-case that `SyncClient` and `SyncService` should be written for. Multi-service seems like an advanced use-case.

PR #100, opened 2017-02-11
fix(bincode): updates to support bincode 1.0.0-alpha2

compressed:
So it looks like bincode 1.0.0-alpha1 isn't compatible with bincode 1.0.0-alpha2.

These changes should fix the breaking build for now.

From what I read in the bincode repo 1.0.0 should come out soon so we can update the dependency accordingly when it does.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -4,7 +4,7 @@
 // This file may not be copied, modified, or distributed except according to those terms.
 
 use {Reactor, WireError};
-use bincode::serde::DeserializeError;
+use bincode;
 use futures::{self, Future};
 use protocol::Proto;
 #[cfg(feature = "tls")]
@@ -18,7 +18,7 @@ use tokio_proto::BindClient as ProtoBindClient;
 use tokio_proto::multiplex::Multiplex;
 use tokio_service::Service;
 
-type WireResponse<Resp, E> = Result<Result<Resp, WireError<E>>, DeserializeError>;
+type WireResponse<Resp, E> = Result<Result<Resp, WireError<E>>, bincode::Error>;
 type ResponseFuture<Req, Resp, E> = futures::Map<<BindClient<Req, Resp, E> as Service>::Future,
                                             fn(WireResponse<Resp, E>) -> Result<Resp, ::Error<E>>>;
 type BindClient<Req, Resp, E> = <Proto<Req, Result<Resp, WireError<E>>> as
@@ -117,7 +117,7 @@ impl<Req, Resp, E> Client<Req, Resp, E>
 
     fn map_err(resp: WireResponse<Resp, E>) -> Result<Resp, ::Error<E>> {
         resp.map(|r| r.map_err(::Error::from))
-            .map_err(::Error::ClientDeserialize)
+            .map_err(::Error::ClientSerialize)
             .and_then(|r| r)
     }
 }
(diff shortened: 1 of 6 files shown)


tikue:
Yeah I think now that derive macros are stabilized, bincode's decided to drop support for rustc_serialize.

src/errors.rs
```diff
@@ -17,11 +16,11 @@ pub enum Error<E> {
     ///
     /// Typically this indicates a faulty implementation of `serde::Serialize` or
     /// `serde::Deserialize`.
-    ClientDeserialize(bincode::serde::DeserializeError),
+    ClientDeserialize(::bincode::Error),
```
tikue: Let's just remove this variant entirely, since the inner types are the same now.

src/client.rs
```diff
@@ -4,7 +4,7 @@
 // This file may not be copied, modified, or distributed except according to those terms.
 
 use {Reactor, WireError};
-use bincode::serde::DeserializeError;
+use bincode::Error as BincodeError;
```
tikue: Nit: `use bincode;` then refer to `bincode::Error` to be consistent with other modules.

src/server.rs
```diff
@@ -4,7 +4,7 @@
 // This file may not be copied, modified, or distributed except according to those terms.
 
 use {REMOTE, Reactor};
-use bincode::serde::DeserializeError;
+use bincode::Error as BincodeError;
```
tikue: Same here, just `use bincode;`

src/errors.rs
compressed: Okay. I changed the comment's wording a bit now since `ClientSerialize` is used for both types of errors.

src/client.rs
compressed: Done.

src/server.rs
compressed: Done

src/errors.rs
tikue: Oh, maybe we should coalesce the server serialization errors as well, for consistency?

src/errors.rs
```diff
@@ -54,8 +48,7 @@ impl<E: StdError + Deserialize + Serialize + Send + 'static> fmt::Display for Er
 impl<E: StdError + Deserialize + Serialize + Send + 'static> StdError for Error<E> {
     fn description(&self) -> &str {
         match *self {
-            Error::ClientDeserialize(_) => "The client failed to deserialize the server response.",
-            Error::ClientSerialize(_) => "The client failed to serialize the request.",
+            Error::ClientSerialize(_) => "The client failed to (de)serialize the request.",
```
tikue: Request or response. So maybe "The client failed to serialize the request or deserialize the response."?

compressed: Makes sense updated.

src/errors.rs
compressed: Yeah I was wondering if you wanted to do that. I think it's good to eliminate as well.

I removed both `Error::ServerDeserialize` and `WireError::ServerDeserialize`.

tikue:
Looks like travis is unhappy

compressed:
Sorry about that. I could have sworn I ran cargo test and it passed... but clearly not 😄 

I fixed the broken test case now.

PR #110, opened 2017-02-19
Prepush script improvement/simplification

shaladdle:
- Remove TARPC_USE_CURRENT_TOOLCHAIN. Things only build on nightly, we might as well just make the pre-push run that way and change it when things work on stable.
- tls and non-tls build/test steps are separate

Examples:

Passing:

<a href="http://i.imgur.com/wGA2ohQ.png">
  
</a>

Failing:

<a href="http://i.imgur.com/3KebX9b.png">
  
</a>


shaladdle:
Also now it prints out the output of the cargo command if it failed.

PR #112, opened 2017-02-21
Change sync listen to return a handle which runs the server

shaladdle:
For now, this handle contains the bound address and a `run` method. This provides a convenient way to block while the server runs/serves, which is generally something that's desirable in a sync server. In the future it'd be good to add a `stop` method, mostly for tests.

In the next PR I'll update examples to be more realistic.

Thanks @tikue for doing the macro changes!

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,9 +76,10 @@ impl SyncService for HelloServer {
 fn main() {
     let (tx, rx) = mpsc::channel();
     thread::spawn(move || {
-        let handle = HelloServer.listen("localhost:0", server::Options::default())
+        let mut handle = HelloServer.listen("localhost:0", server::Options::default())
             .unwrap();
         tx.send(handle.addr()).unwrap();
+        handle.run();
     });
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -72,10 +74,13 @@ impl SyncService for HelloServer {
...
 fn main() {
-    let addr = HelloServer.listen("localhost:0".first_socket_addr(),
-                                  server::Options::default())
-                          .unwrap();
-    let mut client = SyncClient::connect(addr, client::Options::default()).unwrap();
+    let (tx, rx) = mpsc::channel();
+    thread::spawn(move || {
+        let handle = HelloServer.listen("localhost:0", server::Options::default())
+            .unwrap();
+        tx.send(handle.addr()).unwrap();
```
jonhoo: Shouldn't this also have a `handle.run()`?

shaladdle: Done, thanks for catching that.

PR #113, opened 2017-02-21
Return a concrete type from server::listen

tikue:
And change FutureServiceExt::listen to return `(SocketAddr, Listen)`. The macro now generates a new public item called `Listen` that impls `Future<Item=(), Error=()>`.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1034,8 +1034,8 @@ mod functional_test {
     }
 
     mod future {
-        use futures::{Finished, finished};
         use super::{FutureClient, FutureService, env_logger, start_server_with_async_client};
+        use futures::{Finished, finished};
         use tokio_core::reactor;
         use util::Never;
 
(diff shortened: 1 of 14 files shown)


src/macros.rs
```diff
@@ -446,6 +446,43 @@ macro_rules! service {
...
+            type Error = ();
+
+            fn poll(&mut self) -> $crate::futures::Poll<(), ()> {
+                self.inner.poll()
+            }
+        }
+
         /// Provides a function for starting the service. This is a separate trait from
         /// `FutureService` to prevent collisions with the names of RPCs.
         pub trait FutureServiceExt: FutureService {
```
shaladdle: Can you update the comments for this? It says stuff about running on the default tokio loop, which I believe is incorrect now, since the caller has to do the running.

tikue: Done.

PR #114, opened 2017-02-22
Fix throughput example

shaladdle:
The reactor wasn't being run. Also, I get speeds of ~1MB/s on both my macbook and my linux machine, so I reduced the transfer size for the tarpc case to 24MB so it finishes in about 24s, instead of 4 minutes.

### DIFF
--- a/examples/throughput.rs
+++ b/examples/throughput.rs
@@ -18,6 +18,7 @@ extern crate tokio_core;
 use std::io::{Read, Write, stdout};
 use std::net;
 use std::sync::Arc;
+use std::sync::mpsc;
 use std::thread;
 use std::time;
 use tarpc::{client, server};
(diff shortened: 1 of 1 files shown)


tikue:
I'd rather just fix bincode. It's a 10 line patch, so hopefully they'd accept it.

examples/throughput.rs
```diff
@@ -103,5 +111,5 @@ fn main() {
     let _ = env_logger::init();
     let _ = *BUF; // To non-lazily initialize it.
     bench_tcp(256 << 20);
-    bench_tarpc(256 << 20);
+    bench_tarpc(24 << 20);
```
tikue: Keep this 256. I'd rather just wait for bincode to fix their performance issues.

examples/throughput.rs
```diff
@@ -55,12 +56,19 @@ impl FutureService for Server {
...
-    let mut client = SyncClient::connect(addr, client::Options::default()).unwrap();
+    let (tx, rx) = mpsc::channel();
+    thread::spawn(move || {
+        let mut reactor = reactor::Core::new().unwrap();
+        let addr = Server.listen("localhost:0".first_socket_addr(),
+                    &reactor.handle(),
+                    server::Options::default())
+            .unwrap();
+        tx.send(addr).unwrap();
+        loop {
```
tikue: change to `reactor.run(server);`

examples/throughput.rs
```diff
@@ -55,12 +56,19 @@ impl FutureService for Server {
...
-    let reactor = reactor::Core::new().unwrap();
-    let addr = Server.listen("localhost:0".first_socket_addr(),
-                &reactor.handle(),
-                server::Options::default())
-        .unwrap();
-    let mut client = SyncClient::connect(addr, client::Options::default()).unwrap();
+    let (tx, rx) = mpsc::channel();
+    thread::spawn(move || {
+        let mut reactor = reactor::Core::new().unwrap();
+        let addr = Server.listen("localhost:0".first_socket_addr(),
```
tikue: change to `let (addr, server) = ...`

shaladdle: Done

examples/throughput.rs
shaladdle: Done

examples/throughput.rs
shaladdle: Done

TyOverby:
You'll likely get even better perf once https://github.com/TyOverby/bincode/pull/103 lands and you can avoid doing a byte-shuffle on each read.

PR #117, opened 2017-02-23
Add server::Handle::shutdown

### LINKED ISSUE #62: Should server handle expose a fn for graceful shutdown?
One argument against is that the user can stop running their `reactor::Core` whenever they please. However, that would be an all-or-nothing situation. It might still be useful to shutdown some services but not all.

tikue:
Fixes #62 

Edit: as @jonhoo pointed out, this doesn't actually allow `run` to return right now. It just sends servers into lame duck mode, where they honor existing connections but don't accept new ones.

Edit 2: the server now shuts down after all pre-lameduck connections are closed.

Edit 3: changed to two handles: `server::sync::Handle` and `server::future::Handle`. The future one has `addr()` and `shutdown()` and the sync one has those plus `run()`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ description = "An RPC framework for Rust with a focus on ease of use."
 travis-ci = { repository = "google/tarpc" }
 
 [dependencies]
-bincode = { git = "https://github.com/tikue/bincode", branch = "faster-byte-buf-deserialization" }
+bincode = "1.0.0-alpha2"
 byteorder = "1.0"
 cfg-if = "0.1.0"
 futures = "0.1.7"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ description = "An RPC framework for Rust with a focus on ease of use."
 travis-ci = { repository = "google/tarpc" }
 
 [dependencies]
-bincode = "1.0.0-alpha2"
+bincode = { git = "https://github.com/tikue/bincode", branch = "faster-byte-buf-deserialization" }
 byteorder = "1.0"
 cfg-if = "0.1.0"
 futures = "0.1.7"
--- a/src/server.rs
+++ b/src/server.rs
@@ -156,11 +156,21 @@ pub struct Shutdown {
 }
 
 impl Shutdown {
-    fn is_lameduck(&self) -> bool {
+    /// Returns `true` iff the server has entered lameduck mode, in which existing connections
+    /// are honored but no new connections are accepted.
+    ///
+    /// Returns `true` if the server is completely shutdown, as well.
+    pub fn is_lameduck(&self) -> bool {
         self.lameduck.load(Ordering::SeqCst)
     }
 
-    /// Shuts down the server immediately.
+    /// Initiates an orderly server shutdown.
+    ///
+    /// First, the server enters lameduck mode, in which
+    /// existing connections are honored but no new connections are accepted. Then, once all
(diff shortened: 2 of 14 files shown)


src/server.rs
```diff
@@ -140,15 +141,61 @@ pub fn listen<S, Req, Resp, E>(new_service: S,
...
+                        future::Either::A(future::ok(()))
+                    }
+                    None => {
+                        debug!("Shutdown hook dropped; never shutting down.");
+                        future::Either::B(future::empty())
+                    }
+                }
+            });
+        reactor.handle().spawn(server.select(shutdown).then(|_| {
+            debug!("Entering lame duck mode.");
```
jonhoo: What does this mean?
Also, am I reading this right that the server won't *actually* shut down? That is, will the thread that started the server now return from the call to `.run()`?

tikue: It means existing connections will be honored but no new connections will be accepted.

Yeah, so I was a bit hesitant to implement that as part of this, because it'd require one of the following, all of which I wanted to spend a bit more time thinking about and getting feedback on:

1. Have the `Handle` hold the server future so that it can run it directly. Then `run` return once the server stops accepting connections. This requires exposing some type parameters on `Handle`.
2. Same as 1 but box the server to avoid type parameters. Probably not a huge performance hit since, but would be better to benchmark first.
3. Signal shutdown some other way, maybe via an `AtomicBool`. Not sure if there'd be any overhead in checking an atomic once per iteration of `loop { reactor.turn(None); }`.

tikue: I'm going to go with 2. It also means shutdown will be instantaneous rather than honoring existing connections. I think that's fine for its purpose.

Cargo.toml
```diff
@@ -15,7 +15,7 @@ description = "An RPC framework for Rust with a focus on ease of use."
 travis-ci = { repository = "google/tarpc" }
 
 [dependencies]
-bincode = "1.0.0-alpha2"
+bincode = { git = "https://github.com/tikue/bincode", branch = "faster-byte-buf-deserialization" }
```
jonhoo: Was this an intentional inclusion?

tikue: Nope, thanks!

tikue:
The `future::shutdown` test is failing. Not sure why at the moment.

tikue:
Tokio seems to have plans to support shutdown. See https://github.com/tokio-rs/tokio-core/issues/33. I don't think we need to wait for that to play out before making progress on this. We can always revisit later.

src/server.rs
```diff
@@ -119,51 +123,406 @@ impl Options {
...
+    let connections2 = connections.clone();
+
+    let inner = shutdown_rx.take(1)
+        .map(ShutdownSetter { shutdown: shutdown })
+        .merge(connection_rx.map(ConnectionWatcher { connections: connections }))
+        .take_while(ShutdownPredicate {
+            shutdown: shutdown2,
+            connections: connections2,
+        })
+        .map_err(Warn("UnboundedReceiver resolved to an Err; can it do that?"))
```
shaladdle: Is there any guidance from the docs about whether this can happen? If it's never supposed to resolve to an error, maybe this should be `unreachable!`?

tikue: Yeah I just tried replacing the `()` with `!` in `futures-rs` and it still compiles, so it must never be used.

tikue: I'll do it in the next PR (this file doesn't actually exist anymore)

PR #121, opened 2017-02-25
Make SyncClient Send again (and Clone too!).

tikue:
The basic strategy is to start a reactor on a dedicated thread running a request stream. Requests are spawned onto the reactor, allowing multiple requests to be processed concurrently. For example, if you clone the client to make requests from multiple threads, they won't have to wait for each others' requests to complete before theirs start being sent out.

Also, client rpcs only take &self now, which was also required for clients to be usable in a service.

Also added a test to prevent regressions.

This is a followup to #117

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ fn main() {
         tx.send(handle.addr()).unwrap();
         handle.run();
     });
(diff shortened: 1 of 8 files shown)


shaladdle:
Can you do a merge with master please?

tikue:
Yep, on it, had to rebase actually...one min

tikue:
Done.

PR #122, opened 2017-02-26
Change module structure.

### LINKED ISSUE #66: Clean up crate structure
- Which modules should be public?
- Which types should be reexported from the crate root?
- Which types must be public but should be hidden from documentation?
- Which types are currently public but shouldn't be? e.g. `client::Either`, `__tarpc_service_Request`, etc.

### LINKED ISSUE #123: Rename Error::*Serialize to Error::*Deserialize
Serialize errors are swallowed by the `Sink` half of the framed stream; the client just sees a broken pipe. So the names should reflect that.

tikue:
Previous: `client::{sync, future}` and `server::{sync, future}`
Now: `sync::{client, server}` and `future::{client, server}`

The instigation for this is that, when there are separate sync/future versions of `Options`, they'd have to be fully qualified as, e.g., `server::sync::Options`, because `sync::Options` could be client options or server options.

But also, I expect users will usually be using all futures or all sync, and this makes that a bit more ergonomic. You can just do
```rust
use sync::{client, server}
```
rather than

```rust
use client::sync;
use server::sync;
```

This is built off of #117 and #121 because I didn't want to end up dealing with a bunch of merge conflicts.

Fixes #66
Fixes #123


Cargo.toml
```diff
@@ -38,6 +38,7 @@ chrono = "0.3"
 env_logger = "0.3"
 futures-cpupool = "0.1"
 clap = "2.0"
+service-fn = { git = "https://github.com/tokio-rs/service-fn" }
```
shaladdle: I prefer we pin to a specific hash if there's no crate. Is that OK or do you still want to track master?

src/errors.rs
```diff
@@ -84,7 +80,7 @@ impl<E> From<WireError<E>> for Error<E> {
 #[derive(Deserialize, Serialize, Clone, Debug)]
 pub enum WireError<E> {
     /// Error in serializing the server response or deserializing the client request.
```
shaladdle: Is this comment still accurate? If so it's confusing that this can be for a failed deserialize of a response.

Cargo.toml
tikue: Fixed.

src/errors.rs
tikue: Fixed.

PR #128, opened 2017-03-08
Set a default max packet size.

### LINKED ISSUE #127: Server options should include a max packet size
And the default should be sane. A few megs perhaps.

tikue:
Fixes #127

### DIFF
--- a/src/future/client.rs
+++ b/src/future/client.rs
@@ -27,14 +27,40 @@ cfg_if! {
 }
 
 /// Additional options to configure how the client connects and operates.
-#[derive(Default)]
 pub struct Options {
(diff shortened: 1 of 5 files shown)


src/future/client.rs
```diff
@@ -27,14 +27,40 @@ cfg_if! {
...
     reactor: Option<Reactor>,
     #[cfg(feature = "tls")]
     tls_ctx: Option<Context>,
 }
 
+impl Default for Options {
+    #[cfg(feature = "tls")]
+    fn default() -> Self {
+        Options {
+            max_payload_size: 2_000_000,
```
tikue: Done

PR #129, opened 2017-03-08
Thread pool

### LINKED ISSUE #132: warning: use of deprecated item: renamed to `send`
> `warning: use of deprecated item: renamed to `send`
  --> src\main.rs:23:1
   |
23 |   service! {
   |  _^ starting here...
24 | |     rpc simplify(src: String, target: String) -> String;
25 | | }
   | |_^ ...ending here
   |
   = note: #[warn(deprecated)] on by default
   = note: this error originates in a macro outside of the current crate`

i use the code in the example on master branch

tikue:
I also tried to clean up the macro code a bit.

Ok ok ok I got lazy and stuffed a *ton* of stuff in this PR. I'm shameless. You can ask me to try to separate it if desired.

A large portion of this PR is rewriting structs impling the Fn traits to be hand-written futures, allowing for the removal of some unstable features. My hope is to get as close to stable, feature-wise, as possible.

Fixes #132


tikue:
Wew. Merging = complete.

tikue:
Friendly ping :)

src/future/server/shutdown.rs
```diff
@@ -0,0 +1,184 @@
...
+                    None => Ok(Async::Ready(None)),
+                }
+            }
+            Async::Ready(Some(())) => {
+                match self.poll_connections() {
+                    Err(e) => {
+                        self.queued_error = Some(e);
+                        Ok(Async::Ready(Some(())))
+                    }
+                    Ok(Async::NotReady) | Ok(Async::Ready(None)) => {
```
shaladdle: Can these last two be merged more? eg

```rust
Ok(Async::NotReady) | Ok(Async::Ready(None)) | Ok(Async::Ready(Some(())) => Ok(Async::Ready(Some())))
```

src/stream_type.rs
```diff
@@ -52,4 +54,38 @@ impl io::Write for StreamType {
     }
 }
 
-impl Io for StreamType {}
+impl AsyncRead for StreamType {
+    unsafe fn prepare_uninitialized_buffer(&self, buf: &mut [u8]) -> bool {
```
shaladdle: Is there a measurable performance gain that this provides? I read the docs about the functions you're delegating to here and it looks fine. In general though it seems good to justify usage of unsafe in a comment, preferably with perf numbers.

src/future/server/shutdown.rs
tikue: Good catch.

tikue: Relatedly, there was an RFC to allow writing `Ok(Async::Ready(None | Some(())))` but I think it was postponed...

src/stream_type.rs
tikue: Yeah there was a *huge* performance difference. Like, I think throughput doubled or something crazy with this implemented. The problem is that otherwise the message buffer is always zeroed, which matters a lot for large buffer sizes like in the throughput benchmark. The basic idea, which I'll add a comment for, is that by overriding the default for this fn, `StreamType` is obliged to never read the uninitialized buffer (which it doesn't).

tikue: In general no sane implementation of `Read` ever needs to read from the buffer it's writing into, but since safe rust can't assume that, this fn must be unsafe.

PR #163, opened 2017-08-02
Use #[allow(unreachable_patterns)] to simplify some macro code

tikue:
This means we don't need `NotIrrefutable` variants in the Request, Response, and Error enums.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -146,7 +146,6 @@ macro_rules! service {
         #[allow(non_camel_case_types, unused)]
         #[derive($crate::serde_derive::Serialize, $crate::serde_derive::Deserialize)]
         pub enum Request__ {
-            NotIrrefutable(()),
             $(
                 $fn_name{ $($arg: $in_,)* }
             ),*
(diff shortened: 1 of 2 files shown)


src/macros.rs
```diff
@@ -512,23 +510,21 @@ macro_rules! tarpc_service_then__ {
...
-                        msg__
-                    {
-                        ::std::result::Result::Ok(msg__)
-                    } else {
-                       unreachable!()
+                    #[allow(unreachable_patterns)]
+                    match msg__ {
+                        Response__::$fn_name(msg__) =>
+                            ::std::result::Result::Ok(msg__),
+                        _ => unreachable!(),
```
shaladdle: Can this code path be hit by an adversarial client sending bad requests to the server? It would cause the server to crash, right? Or maybe that would just cause a deserialization error?

I'm approving anyways because it looks like this these `unreachable`s were there before.

tikue: This code is client-side. This code path can be hit by a non-tarpc server implementing the service protocol incorrectly. So right now there's kind of the assumption that the only tarpc protocol-serving servers are running tarpc code.

PR #175, opened 2017-12-04
Updates

imp:
Pull up dependencies and some assorted clippy warning cleanups

### DIFF
--- a/src/future/client.rs
+++ b/src/future/client.rs
@@ -90,17 +90,18 @@ enum Reactor {
 
 impl fmt::Debug for Reactor {
     fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
-        const HANDLE: &'static &str = &"Reactor::Handle";
-        const HANDLE_INNER: &'static &str = &"Handle { .. }";
-        const REMOTE: &'static &str = &"Reactor::Remote";
-        const REMOTE_INNER: &'static &str = &"Remote { .. }";
+        const HANDLE: &str = "Reactor::Handle";
+        const HANDLE_INNER: &str = "Handle { .. }";
+        const REMOTE: &str = "Reactor::Remote";
+        const REMOTE_INNER: &str = "Remote { .. }";
 
         match *self {
-            Reactor::Handle(_) => f.debug_tuple(HANDLE).field(HANDLE_INNER).finish(),
-            Reactor::Remote(_) => f.debug_tuple(REMOTE).field(REMOTE_INNER).finish(),
+            Reactor::Handle(_) => f.debug_tuple(HANDLE).field(&HANDLE_INNER).finish(),
+            Reactor::Remote(_) => f.debug_tuple(REMOTE).field(&REMOTE_INNER).finish(),
         }
     }
 }
+
 #[doc(hidden)]
 pub struct Client<Req, Resp, E>
 where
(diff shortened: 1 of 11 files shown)


src/future/client.rs
```diff
@@ -90,10 +90,10 @@ enum Reactor {
 
 impl fmt::Debug for Reactor {
     fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
-        const HANDLE: &'static &'static str = &"Reactor::Handle";
-        const HANDLE_INNER: &'static &'static str = &"Handle { .. }";
-        const REMOTE: &'static &'static str = &"Reactor::Remote";
-        const REMOTE_INNER: &'static &'static str = &"Remote { .. }";
+        const HANDLE: &'static &str = &"Reactor::Handle";
```
tikue: Can this outer 'static also be elided? Likewise below.

imp: Right. Not sure why I left it... Pushed another commit that fixes it.
However, why not derive `Debug` instead of implementing it manually?

tikue: Not all of the fields of Reactor had impl'd `Debug` when this was written. If that's no longer the case, then this should absolutely be deriving `Debug`. Feel free to submit another PR with that change. In the meantime, I'll merge this in!

src/future/server/mod.rs
```diff
@@ -127,7 +127,7 @@ impl fmt::Debug for Acceptor {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
         use self::Acceptor::*;
         #[cfg(feature = "tls")]
-        const TLS: &'static &'static str = &"TlsAcceptor { .. }";
+        const TLS: &'static &str = &"TlsAcceptor { .. }";
```
tikue: A few more `'static`s

src/future/server/mod.rs
```diff
@@ -223,9 +223,9 @@ impl Options {
 impl fmt::Debug for Options {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
         #[cfg(feature = "tls")]
-        const SOME: &'static &'static str = &"Some(_)";
+        const SOME: &'static &str = &"Some(_)";
```
tikue: A few more `'static`s

src/future/server/mod.rs
```diff
@@ -223,9 +223,9 @@ impl Options {
...
         #[cfg(feature = "tls")]
-        const NONE: &'static &'static str = &"None";
+        const NONE: &'static &str = &"None";
```
tikue: A few more `'static`s

src/future/server/mod.rs
```diff
@@ -352,7 +352,7 @@ where
     St: fmt::Debug,
 {
     fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
-        const HANDLE: &'static &'static str = &"Handle { .. }";
+        const HANDLE: &'static &str = &"Handle { .. }";
```
tikue: A few more `'static`s

src/future/client.rs
imp: Cool, thanks!

src/future/server/mod.rs
imp: Well, a few less `'static`s now :)

tikue:
Thanks @imp !

PR #180, opened 2018-03-26
Update deps

tikue:


examples/concurrency.rs
```diff
@@ -147,7 +147,7 @@ fn run_once(
 }
 
 fn main() {
-    let _ = env_logger::init();
+    env_logger::init();
```
shaladdle: Why init here and try_init above?

tikue: init() will panic if called more than once, whereas `try_init()` returns a result.. In unit tests, each `#[test]` needs to init the logger, because we don't want to rely on an order of test execution. Therefore, tests need to use `try_init`, but examples/ files all have a single main method, so it's safe to call `init()`.

PR #208, opened 2018-12-09
Update regex to match diffs output by cargo fmt.

martica:
It appears the header of the diffs output by cargo fmt have changed. It now says "Diff in /blah/blah/blah.rs at line 99:" Matching on lines starting with + or - should be more future-proof against changes to the surroundings.


tikue:
Thanks!

martica:
Cheers. Thanks for having an easy to find hook configuration to borrow. :)

PR #216, opened 2019-03-28
Port to futures-preview 0.3.0-alpha.14

vorot93:


tikue:
Sweet, thanks for spearheading this! What are the major changes? Looks like Sink is now generic over the item rather than it being an associated type.

vorot93:
@tikue futures-preview now has its own sink compat shims. I had to resort to extremely fragile transmute to expose a private method though. Also `CloneSpawn` is broken and has been replaced by a mutex in `SEED_SPAWN`.

On a side note, do we really need `tarpc::init()`? Is it not better to use the executor upon which tarpc client/server future is spawned?

tikue:
Regarding `tarpc::init()`, I'd prefer to track that in a dedicated bug, but briefly, I'm not opposed to having a config option in `client::Config` and `server::Config` to set the executor, but I think it's nice to have a default specified by `init()`.

> Is it not better to use the executor upon which tarpc client/server future is spawned?

I'm pretty sure this is possible! `init()` just takes something `Spawn + Clone`, e.g. `DefaultExecutor::current().compat()`. Nothing stops the user from using this same executor elsewhere. In fact doesn't `tokio::run()` use the current executor too?

Nemo157:
@vorot93 I can't remember how I ended up looking at this PR, but if you want to open an issue about the missing method on `Compat01As03Sink` I'm sure it could get added easily (by the looks of it you just need `fn as_ref(&self) -> &S`?)

tikue:
@Nemo157 that's a great suggestion! @vorot93 do you want to make that PR?

vorot93:
@Nemo157 @tikue rust-lang-nursery/futures-rs#1524

vorot93:
@tikue transmute is no longer there

rpc/src/lib.rs
```diff
@@ -149,9 +155,9 @@ impl<T> Request<T> {
 pub(crate) type PollIo<T> = Poll<Option<io::Result<T>>>;
 
 static INIT: Once = Once::new();
-static mut SEED_SPAWN: Option<Box<dyn CloneSpawn>> = None;
+static mut SEED_SPAWN: Option<Arc<Mutex<dyn Spawn>>> = None;
```
tikue: This is the only thing I want to look into further before merging. I will do my best to investigate tonight!

rpc/src/transport/mod.rs
```diff
@@ -23,7 +24,7 @@ pub mod channel;
 pub trait Transport
 where
     Self: Stream<Item = io::Result<<Self as Transport>::Item>>,
-    Self: Sink<SinkItem = <Self as Transport>::SinkItem, SinkError = io::Error>,
+    Self: Sink<<Self as Transport>::SinkItem, SinkError = io::Error>,
```
tikue: Unrelated thought: now that Sink is generic over SinkItem, it probably makes sense to make Transport generic over it as well!

I'd accept a separate PR making that change.

PR #230, opened 2019-05-10
Update to futures-preview 0.3.0-alpha.16

vorot93:
This includes a switch to new `await` syntax. Warrants a version bump as well.


tikue:
Thanks so much! Do you know if the `generators` feature is still needed?

tarpc/examples/server_calling_server.rs
```diff
@@ -59,7 +58,7 @@ impl DoubleService for DoubleServer {
 
     fn double(self, _: context::Context, x: i32) -> Self::DoubleFut {
         async fn double(mut client: add::Client, x: i32) -> Result<i32, String> {
-            let result = await!(client.add(context::current(), x, x));
+            let result = client.add(context::current(), x, x).await;
```
tikue: No reason for a temp variable anymore; mind inlining this line with the next? It's a good example of await chaining.

vorot93: Good catch!

vorot93:
`generators` feature is not necessary any more. Removed it.

Issue #217, opened 2019-03-29
Rpc session on user-provided connection


elbaro:
Can one create a rpc client/server on a user provided connection (e.g. std::net or tokio TcpStream/UnixStream)? I would like to have more control on connections and do these things:

- multiplexing, serve different RPC servers
```rs
let stream = TcpStream::connect("127.0.0.1:555".parse().unwrap()).unwrap();
let plex = some_multiplexer::new(stream);
let conn1 = plex.open();
let conn2 = plex.open();
let rpc_client1 = RpcClient1::new(conn1);
let rpc_client2 = RpcClient2::new(conn2);
```

- Gracious exit with stream/future
```rs
use stream_cancel::StreamExt;

let listener = tokio::net::UnixListener::bind("/tmp/example.sock").unwrap();
listener
	.incoming()
	.take_until(is_terminated)
	.for_each(move |stream| {
		let stream = stream.take_until(is_terminated);
		let rpc_server = RpcServer::new(stream)
                rpc_server.serve_calls();
	});

exit(); // is_terminate resolves to completed
```

- Built-in policy like ConnectionFilter can be customized

tikue:
Hi! I think you should be able to do some of these things today.

For multiplexing, you could either provide a multiplexing transport, possibly by multiplexing across multiple [channel transports](https://github.com/google/tarpc/blob/master/rpc/src/transport/channel.rs); or you could implement a service that multiplexes across other services, like I do in [service registry](https://github.com/google/tarpc/blob/4e0be5b626dc940e2c4b9dd61e87ab0302113938/tarpc/examples/service_registry.rs#L377).

For gracious exit, you can use the server's connection stream to decide when to stop. I do something like that in the [readme example](https://github.com/google/tarpc/blob/4e0be5b626dc940e2c4b9dd61e87ab0302113938/tarpc/examples/readme.rs#L60). For closing individual connections, I believe you can select the future representing the connection with your exit signal. So instead of `incoming.respond_with(serve(server))` you can do `incoming.for_each(|channel| select(
    channel.respond_with(serve(server.clone())), 
    exit_signal
))`.

Can you say more about how you'd like to customized ConnectionFilter? That's a built-in policy, like you say, but you can provide your own filters on the connection stream, similar to how you'd hook in gracious exit.

elbaro:
Thanks for the answer. Basically I want lower-level channel API rather than `Server` struct.

[...]

PR #233, opened 2019-05-12
Make channel functionality opt-in

tikue:
## Connection Limits

The problem with having ConnectionFilter default-enabled is elaborated on in https://github.com/google/tarpc/issues/217. The gist of it is not all servers want a policy based on `SocketAddr`. This PR allows customizing the behavior of ConnectionFilter, at the cost of not having it enabled by default. However, enabling it is as simple as one line:

```rust
incoming.max_channels_per_key(10, ip_addr)
```

The second argument is a key function that takes the user-chosen transport and returns some hashable, equatable, cloneable key. In the above example, it returns an `IpAddr`.

This also allows the `Transport` trait to have the addr fns removed, which means it has become simply an alias for `Stream + Sink`.

## Per-Channel Request Throttling

With respect to Channel's throttling behavior, the same argument applies. There isn't a one size fits all solution to throttling requests, and the policy applied by tarpc is just one of potentially many solutions. As such, `Channel` is now a trait that offers a few combinators, one of which is throttling:

```rust
channel.max_concurrent_requests(10).respond_with(serve(Server))
```

This functionality is also available on the existing `Handler` trait, which applies it to all incoming channels and can be used in tandem with connection limits:

```rust
incoming
    .max_channels_per_key(10, ip_addr)
    .max_concurrent_requests_per_channel(10).respond_with(serve(Server))
```

## Global Request Throttling

I've entirely removed the overall request limit enforced across all channels. This functionality is easily gotten back via [`StreamExt::buffer_unordered`](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.1/futures/stream/trait.StreamExt.html#method.buffer_unordered), with the difference being that the previous behavior allowed you to spawn channels onto different threads, whereas `buffer_unordered ` means the `Channels` are handled on a single thread (the per-request handlers are still spawned). Considering the existing options, I don't believe that the benefit provided by this functionality held its own.


tikue:
I'm going to merge this soon. My main concern is that this is a rather large, *silent*, breaking change: users' existing code will continue to work if using the default server config because they won't notice the removed config fields.

My hope is that semantic versioning will prevent users of receiving these changes accidentally. I'll also include a detailed changelog with the next release.

tikue:
Closing in lieu of a PR from my personal remote.

PR #245, opened 2019-07-23
Use runtime crate everywhere

### LINKED ISSUE #243: Use runtime crate in lieu of init fn
https://github.com/rustasync/runtime

The `runtime` crate provides a runtime-agnostic attach point for async libraries. Importantly, it provides a `spawn` function exactly like the one tarpc uses internally.

vorot93:
Enables usage of executor from `runtime` family of crates unless `tarpc::init` is called. Updated examples and tests to use `runtime` and async `main()`.

Fixes #243


rpc/src/client/channel.rs
```diff
@@ -250,7 +250,7 @@ where
     let (cancellation, canceled_requests) = cancellations();
     let canceled_requests = canceled_requests.fuse();
 
-    crate::spawn(
+    runtime::spawn(
```
tikue: I'd like to keep `crate::spawn`, here and elsewhere. `runtime::spawn` panics on error and also sets up a channel for the response, neither of which we want.

tarpc/examples/readme.rs
```diff
@@ -77,13 +76,9 @@ async fn run() -> io::Result<()> {
...
-fn main() {
-    tarpc::init(tokio::executor::DefaultExecutor::current().compat());
-
-    tokio::run(
-        run()
-            .map_err(|e| eprintln!("Oh no: {}", e))
-            .boxed()
-            .compat(),
-    );
+#[runtime::main]
```
tikue: I'm surprised this doesn't need to be `runtime::main(Tokio)`? Because `bincode_transport` uses tokio tcp.

tarpc/examples/service_registry.rs
```diff
@@ -406,7 +405,7 @@ async fn run() -> io::Result<()> {
     Ok(())
 }
 
-fn main() {
-    tarpc::init(tokio::executor::DefaultExecutor::current().compat());
-    tokio::run(run().boxed().map_err(|e| panic!(e)).boxed().compat());
+#[runtime::main]
+async fn main() {
+    run().await.unwrap()
```
tikue: You can just change `run` to be named `main` and remove this main entirely

vorot93:
After the initial plunge that broke all test cases, I started from scratch taking into account comments by @tikue.

tikue:
@vorot93 is this ready for review? (Click the button if so)

rpc/src/transport/channel.rs
```diff
@@ -113,33 +112,25 @@ mod tests {
             Ok::<_, io::Error>((response1, response2))
         };
 
-        let (response1, response2) = run_future(future::join(
+        let (response1, response2) = runtime::task::spawn(future::join(
```
tikue: The `join` here was only to make it easier to run the future. Since we're in an async fun, you can just spawn the server first and then await each response individually, like we do in other tests.

rpc/src/transport/channel.rs
```diff
@@ -113,33 +112,25 @@ mod tests {
...
         ))
+        .await
         .1;
 
         trace!("response1: {:?}, response2: {:?}", response1, response2);
 
-        assert!(response1.is_ok());
-        assert_eq!(response1.ok().unwrap(), 123);
+        let response1 = response1.map_err(|e| e.to_string());
+        ensure!(
```
tikue: Let's use `assert_matches` instead? All of this ugly code I wrote before knowing about `assert_matches`.

rpc/src/lib.rs
```diff
@@ -141,7 +141,11 @@ pub fn init(spawn: impl Spawn + Clone + Send + Sync + 'static) {
 }
 
 pub(crate) fn spawn(future: impl Future<Output = ()> + Send + 'static) -> Result<(), SpawnError> {
-    SPAWN.with(|spawn| spawn.borrow_mut().spawn(future))
+    if SEED_SPAWN.get().is_some() {
```
tikue: I actually am in favor of getting rid of SEED_SPAWN and init entirely; sorry if that wasn't clear! The `spawn_boxed` looks good and we can just always use that line.

rpc/src/transport/channel.rs
vorot93: I am afraid that with assert failures will always output `the main future has panicked: Canceled`. Runtime makes us write tests that actually return an error - therefore we should use `ensure!` instead. In fact I am thinking of creating `ensure_eq!` and `ensure_matches!` and proposing them for inclusion into `failure` crate.

rpc/src/lib.rs
vorot93: Sure we can, but that will make `runtime` mandatory for everyone who wants to use `tarpc`. Whether that is desirable (or not) is a matter of discussion, so I've left this safety hatch for now.

tikue: Ah, fair point! Let's leave it for now. I have a branch wherein I make all usage of the global executor opt-in, to facilitate the removal of mandatory `Send + 'static` bounds. We can revisit this then.

rpc/src/transport/channel.rs
tikue: I have a branch where I rewrote this test, and it seems to work as expected. In that branch, I spawn the server and make the client calls directly, not in an async block. I think the problem here is that runtime::spawn() returns a channel Receiver, which just sees a canceled sender.

Here's the code where I rewrote it: https://github.com/tikue/tarpc/blob/runtime/rpc/src/transport/channel.rs#L91-L118

vorot93: I took your version of the code and it spews the same nondescriptive `the main future has panicked: Canceled` when an assert fails.

tikue: That's weird. Did you try checking out my branch? I changed Ok(123) to Ok(124) and got a reasonably nice assertion panic.

vorot93: I have adapted `assert_matches` to the standard propagation semantic.

https://github.com/vorot93/ensure_matches

tikue:
This looks good to me. Is there a reason it's marked as a Draft? Can I mark it "Ready for review" and approve & merge?

tikue:
One thought: we could put the runtime integration behind a default-enabled feature. That way, folks who are just calling `tarpc::init()` don't transitively pull in the `runtime` crates.

(I can also do that in a follow-up commit if you want to submit this as-is.)

vorot93:
I just wanted to port more tests to runtime and make use of `ensure_matches!`, but since my timetable has gone haywire it's better to do it in a separate PR.

vorot93:
Enabling tarpc::init if runtime integration is disabled is a very cool idea! Let's merge this PR as is now though.

tikue:
Thanks for taking the lead on this!

PR #248, opened 2019-07-30
Move `fn serve` to service trait and `fn new_stub` to Client impl

tikue:
With this change, the service definitions don't need to be isolated in their own modules.

Given:

```rust
#[tarpc::service]
trait World { ... }
```

Before this would generate the following items
------
- `trait World`
- `fn serve`
- `struct Client`
- `fn new_stub`

`// Implementation details below`
- `enum Request`
- `enum Response`
- `enum ResponseFut`

And now these items
------
- `trait World {    ...    fn serve }`
- `struct WorldClient ... impl WorldClient {    ...    async fn new }`

`// Implementation details below`
- `enum WorldRequest`
- `enum WorldResponse`
- `enum WorldResponseFut`
- `struct ServeWorld` (new manual closure impl because you can't use impl Trait in trait fns)
```


vorot93:
I feel uneasy about this change. It is more ergonomic and less surprising to the end user when autogen result names are consistent and placed into their own namespace.

tikue:
That's fair. I'll leave this open for a bit in case you or others come to feel differently. I will say that this is the way grpc works.

What do you think of this alternative: doing this just for the client and hidden items, and then moving `serve` and `new_stub` into impls on their respective types? It provides the same isolation benefit while only changing the name of the client. The downside is that it would prevent users from adding their own rpcs named `serve` or `new`, but now that the expansion is done by a proc macro, I can detect that case and provide a good error message (e.g. "RPC name conflicts with generated method `serve`). I could even allow for an attribute meta item that overrides the generated method names in case one wants to have an RPC named `serve` or `new`, and the error message could offer a hint about it.

tikue:
I've made the alternative changes and updated the description. What do you think now? From the user's perspective, the names of the Client and Service are consistent with each other, and the names of the generated functions are consistent across service definitions.

tikue:
@vorot93 any thoughts on the latest changes? To me, it feels much better having the `serve` and `stub` fns as inherent methods on their respective types.

vorot93:
`serve` being an inherent method is most definitely welcome. More consistent naming is too. I still prefer the older namespace option, however at this point it may be just a personal matter of taste.

tikue:
Thanks for the feedback! I'm going to merge this now. However, I want to clarify that I don't think it changes anything fundamental about the namespacing options. You can still out each service definition in its own module and reexport from there.

Issue #212, opened 2019-03-15
Send bound should be optional


vorot93:
Users should be able to opt out of `Send` bound if they are building a single-threaded application (e.g. using `tokio::runtime::current_thread`).

tikue:
I agree this would be useful! I suspect it would require quite a significant refactoring to be able to add this type of customization. `Send` assumptions are baked in pretty heavily at the moment.

PR #250, opened 2019-08-05
Don't bake in Send + 'static.

tikue:
`Send + 'static` was baked in to make it possible to spawn futures onto the default executor. We can accomplish the same thing by offering helper fns that do the spawning while not requiring it for the rest of the functionality.

The major changes are:

- `Client::new` returns a `NewClient` struct that contains the client and its dispatch future (which needs to be driven for progress to be made). It has a helper function, `spawn()`, to achieve the previous functionality: it spawns the dispatch on the default executor and returns the client.
- `Channel::respond_with` returns a `Stream` of futures, each representing a single request's handler. There is now an additional helper method on the returned type, `execute()`, which converts the stream to a single `Future` by spawning all request handler on the default executor and resolving when the channel is closed.

My general feeling is that the ergonomic hit is miniscule, while providing substantially more flexibility to users. Check out the new test in [tarpc/tests/service_functional.rs](https://github.com/tikue/tarpc/blob/remove-send-static/tarpc/tests/service_functional.rs#L135-L191) demonstrating the ability to have an in-process, single-threaded client-server connection that can send unserialized `Rc` objects to each other! Pretty neat.

Fixes https://github.com/google/tarpc/issues/212


vorot93:
One thing is that this PR moves tarpc away from "executor-executable separation" idiom. Ideally tarpc should always give the user a future, awaiting on which would start the client or server. It is a separate topic for discussion though.

tikue:
> One thing is that this PR moves tarpc away from "executor-executable separation" idiom. Ideally tarpc should always give the user a future, awaiting on which would start the client or server. It is a separate topic for discussion though.

This sounds like a good topic to discuss further; do you want to file an issue?

PR #264, opened 2019-08-29
Tokio 0.2.0 alpha.3 and minor fixes

vorot93:
Also took the liberty to cut the last feature flag so that we are 100% working on stable once async/await lands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ your `Cargo.toml`:
 
 ```toml
 futures-preview = "0.3.0-alpha.18"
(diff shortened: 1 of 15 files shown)


vorot93:
0.3.0-alpha.3 has broken hyper, so we should wait for 0.3.0-alpha.4

tikue:
Awesome, thank you!

tikue:
Oops, just saw your second comment. I think it's okay as long as we don't publish a release to crates.io until .4 is out.

PR #273, opened 2019-10-09
Port to pin-project

### LINKED ISSUE #270: Use pin-project in lieu of pin-utils
https://crates.io/crates/pin-project

It appears to be a safer version.

vorot93:
Fixes #270

### DIFF
--- a/rpc/src/client/channel.rs
+++ b/rpc/src/client/channel.rs
@@ -19,10 +19,9 @@ use futures::{
     Poll,
 };
 use log::{debug, info, trace};
-use pin_utils::{unsafe_pinned, unsafe_unpinned};
+use pin_project::{pin_project, pinned_drop};
 use std::{
     io,
-    marker::Unpin,
     pin::Pin,
     sync::{
         atomic::{AtomicU64, Ordering},
@@ -55,9 +54,11 @@ impl<Req, Resp> Clone for Channel<Req, Resp> {
 }
 
 /// A future returned by [`Channel::send`] that resolves to a server response.
+#[pin_project]
 #[derive(Debug)]
 #[must_use = "futures do nothing unless polled"]
 struct Send<'a, Req, Resp> {
+    #[pin]
     fut: MapOkDispatchResponse<SendMapErrConnectionReset<'a, Req, Resp>, Resp>,
 }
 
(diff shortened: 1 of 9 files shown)


rpc/src/client/channel.rs
```diff
@@ -155,19 +139,17 @@ impl<Req, Resp> Channel<Req, Resp> {
 
 /// A server response that is completed by request dispatch when the corresponding response
 /// arrives off the wire.
+#[pin_project(PinnedDrop)]
 #[derive(Debug)]
 struct DispatchResponse<Resp> {
     response: Timeout<oneshot::Receiver<Response<Resp>>>,
+    #[pin]
     ctx: context::Context,
```
tikue: I suspect I did this wrong originally. `context::Context` impls `Unpin` right? We may be able to leave this out.

rpc/src/client/channel.rs
```diff
@@ -246,15 +230,20 @@ where
...
     transport: Fuse<C>,
     /// Requests waiting to be written to the wire.
+    #[pin]
     pending_requests: Fuse<mpsc::Receiver<DispatchRequest<Req, Resp>>>,
     /// Requests that were dropped.
+    #[pin]
     canceled_requests: Fuse<CanceledRequests>,
     /// Requests already written to the wire that haven't yet received responses.
+    #[pin]
     in_flight_requests: FnvHashMap<u64, InFlightData<Resp>>,
```
tikue: Is this another pin that I did incorrectly originally?

rpc/src/server/mod.rs
```diff
@@ -165,6 +165,7 @@ where
 }
 
 /// BaseChannel lifts a Transport to a Channel by tracking in-flight requests.
+#[pin_project]
```
tikue: Can the field `transport` be annotated with `#[pin]` instead of the manual fn below?

rpc/src/server/testing.rs
```diff
@@ -4,57 +4,58 @@ use fnv::FnvHashSet;
...
 use std::io;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 use std::time::SystemTime;
 
+#[pin_project]
 pub(crate) struct FakeChannel<In, Out> {
+    #[pin]
     pub stream: VecDeque<In>,
+    #[pin]
```
tikue: Do these `VecDeque`s actually need to be pinned or do they impl unpin?

vorot93: That would require the inner type to be `Unpin`.

rpc/src/server/mod.rs
vorot93: Done.

PR #278, opened 2019-11-10
multithread example service + remove unnecessary template from codegen

GallagherCommaJack:
there are two different pieces here:

1) The example service was showing a way of starting the server that's probably ill-advised - in particular, using `buffer_unordered` instead of `tokio::spawn` will lead to much worse performance. I also modified the example client to spin up a pool of connections and properly load test the service - if this isn't desirable (it might not be), I can put it behind an optional argument (`--load`?), and go back to the previous behavior when no argument is given.
2) I changed the codegen to emit less... bizarre Rust. In particular, I removed the templating over `C = tarpc::client::Transport` in the client, because when there's an equality constraint that's not actually a generic...


GallagherCommaJack:
@googlebot I signed it!

tikue:
Thanks for making the example better!

Are you sure the codegen change doesn't break the integration with the client trait's provided combinators? See [here](https://github.com/google/tarpc/blob/5f6c3d7d989b7d7267c19d4b89c658450e86614e/rpc/src/client/mod.rs#L34).

GallagherCommaJack:
Uh how would it break exactly? The template was constraining the type parameter `C` to have a specific value, those combinators all still typecheck, and the tests all still pass.

GallagherCommaJack:
After a bit of investigation, I suspect those methods were already broken - client codegen had a `C = Channel<$req, $resp>` constraint, rather than something like  `C : for<'a> Client<'a, Req>`. I'll see if I can fix that.

GallagherCommaJack:
Huh looks like it did work previously - well, I modified the templating to be more straightforward at least - I've literally never seen the `C = ...` notation in the wild before, and I'm not super sure what constraints it was adding.

tikue:
The `<C = Foo>` notation isn't a constraint; it's a default. It allows a user to leave out the type parameter if they wish to use the default. In this case, users can say `Client` instead of `Client<Channel<Request, Response>>`. You can see real examples in the failing [Travis CI tests](https://travis-ci.org/google/tarpc/jobs/609880881?utm_medium=notification&utm_source=github_status):

```rust
error[E0107]: wrong number of type arguments: expected 1, found 0
  --> tarpc/examples/server_calling_server.rs:54:37
   |
54 |         async fn double(mut client: add::AddClient, x: i32) -> Result<i32, String> {
   |                                     ^^^^^^^^^^^^^^ expected 1 type argument
```

GallagherCommaJack:
Ah yep that makes sense - should probably add a test that uses the generic at some point.
I've added a commit that reverts the plugin, so if there's a squashed merge it'll only have the example service update.

example-service/Cargo.toml
```diff
@@ -13,7 +13,7 @@ readme = "../README.md"
 description = "An example server built on tarpc."
 
 [dependencies]
-tarpc-json-transport = { version = "0.1", path = "../json-transport" }
+tarpc-bincode-transport = { version = "0.7", path = "../bincode-transport" }
```
tikue: Any particular reason to use bincode over json? The nice thing about json for examples is that the protocol is human-readable, so easier to debug/play around with when starting out.

example-service/src/client.rs
```diff
@@ -38,20 +39,91 @@ async fn main() -> io::Result<()> {
         .parse::<SocketAddr>()
         .unwrap_or_else(|e| panic!(r#"--server_addr value "{}" invalid: {}"#, server_addr, e));
 
-    let name = flags.value_of("name").unwrap().into();
+    let num_connections: usize = flags
```
tikue: I like the idea of putting the load test behind a flag. That will keep the first experience simpler. Perhaps there can be a server-side load test flag as well, which switches the transport from json to bincode?

example-service/src/client.rs
```diff
@@ -38,20 +39,91 @@ async fn main() -> io::Result<()> {
...
+    ) = (0..num_connections)
+        .map(|_| tokio::sync::mpsc::channel(ROUNDS + 1))
+        .unzip();
+
+    for mut sender in senders.into_iter() {
+        tokio::spawn(async move {
+            sender
+                .send(Instant::now())
+                .await
+                .expect("failed to send instant");
```
tikue: The expect message is only printed in case of failure, so I think the "failed to" prefix is redundant. Also, instead of "instant", perhaps include the purpose of the thing being sent (start time)?

```suggestion
                .expect("send start connecting time");
```

example-service/src/client.rs
```diff
@@ -38,20 +39,91 @@ async fn main() -> io::Result<()> {
...
+            let transport = tarpc_bincode_transport::connect(&server_addr)
+                .await
+                .expect("failed to connect");
+            sender
+                .send(Instant::now())
+                .await
+                .expect("failed to send instant");
```
tikue:
```suggestion
                .expect("send connected time");
```

example-service/src/client.rs
```diff
@@ -38,20 +39,91 @@ async fn main() -> io::Result<()> {
...
+
+            let mut client = service::WorldClient::new(client::Config::default(), transport)
+                .spawn()
+                .expect("failed to spawn client");
+
+            for _ in 0..ROUNDS {
+                client
+                    .hello(context::current(), NAME.into())
+                    .await
+                    .expect("failed to run hello");
```
tikue:
```suggestion
                    .expect("rpc hello");
```

example-service/src/client.rs
```diff
@@ -38,20 +39,91 @@ async fn main() -> io::Result<()> {
...
+                .await
+                .expect("failed to connect");
+            sender
+                .send(Instant::now())
+                .await
+                .expect("failed to send instant");
+
+            let mut client = service::WorldClient::new(client::Config::default(), transport)
+                .spawn()
+                .expect("failed to spawn client");
```
tikue:
```suggestion
                .expect("spawn client");
```

example-service/src/client.rs
```diff
@@ -38,20 +39,91 @@ async fn main() -> io::Result<()> {
...
+
+            for _ in 0..ROUNDS {
+                client
+                    .hello(context::current(), NAME.into())
+                    .await
+                    .expect("failed to run hello");
+                sender
+                    .send(Instant::now())
+                    .await
+                    .expect("failed to send instant");
```
tikue:
```suggestion
                    .expect("send rpc response time");
```

example-service/src/client.rs
```diff
@@ -38,20 +39,91 @@ async fn main() -> io::Result<()> {
...
+            }
+        });
+    }
+
+    let mut latencies = Vec::new();
+    let mut instants = Vec::new();
+    for r in receivers.into_iter() {
+        let ivec: Vec<_> = r.collect().await;
+        for i in ivec.iter() {
```
tikue:
```suggestion
        for i in &ivec {
```

example-service/src/client.rs
```diff
@@ -38,20 +39,91 @@ async fn main() -> io::Result<()> {
...
+                    .send(Instant::now())
+                    .await
+                    .expect("failed to send instant");
+            }
+        });
+    }
+
+    let mut latencies = Vec::new();
+    let mut instants = Vec::new();
+    for r in receivers.into_iter() {
```
tikue:
```suggestion
    for r in receivers {
```

example-service/src/client.rs
```diff
@@ -38,20 +39,91 @@ async fn main() -> io::Result<()> {
...
+        let ivec: Vec<_> = r.collect().await;
+        for i in ivec.iter() {
+            instants.push(*i);
+        }
+        for (i1, i2) in ivec.iter().zip(&ivec[1..]) {
+            let dur = i2.duration_since(*i1);
+            latencies.push(dur);
```
tikue: Am I correct that `latencies` contains the request times as well as the connect time? Should the connect times to be left out?

example-service/src/server.rs
```diff
@@ -62,27 +62,29 @@ async fn main() -> io::Result<()> {
...
+            config.pending_response_buffer = 256;
+            server::BaseChannel::new(config, t)
         })
-        // Max 10 channels.
-        .buffer_unordered(10)
-        .for_each(|_| async {})
-        .await;
+        // limit connections to 1 per IP
+        // disabled to allow for load testing
+        // .max_channels_per_key(1, |t| t.as_ref().peer_addr().unwrap().ip())
```
tikue: I'd just remove this entirely rather than leave it commented out.

example-service/src/client.rs
GallagherCommaJack: Alternately we could just have the server listen for both on different ports, with different settings.

tikue: I like that idea too!

tikue:
@GallagherCommaJack hey! You still planning to pick this up, or should I close it for now?

GallagherCommaJack:
I'm not planning to pick it up anytime soon - changes you suggested all make sense but are asking for more work that doesn't sound particularly fun to do on my end. If I ever get around to them I'll reopen.

PR #280, opened 2019-11-26
Unite most of tarpc into a single crate

vorot93:
This follows in the footsteps of Tokio. It should make maintenance a lot easier while retaining the flexibility that users may want.


README.md
```diff
@@ -41,13 +41,13 @@ Some other features of tarpc:
   sends a request to another server, that server will see an 8s deadline.
 - Serde serialization: enabling the `serde1` Cargo feature will make service requests and
   responses `Serialize + Deserialize`. It's entirely optional, though: in-memory transports can
-  be used, as well, so the price of eerialization doesn't have to be paid when it's not needed.
+  be used, as well, so the price of serialization doesn't have to be paid when it's not needed.
```
tikue: I thought I fixed this, but I only changed the crate documentation :)

tarpc/Cargo.toml
```diff
@@ -13,28 +13,49 @@ readme = "../README.md"
...
+pin-project = "0.4"
+raii-counter = "0.2"
+rand = "0.7"
+tokio-timer = "0.3.0-alpha"
+serde = { optional = true, version = "1.0", features = ["derive"] }
+tokio = { optional = true, version = "0.2.0-alpha" }
+tarpc-plugins = { path = "../plugins" }
+
+async-bincode = { optional = true, version = "0.4" }
+futures-legacy = { optional = true, version = "0.1", package = "futures" }
```
tikue: What still uses this?

tarpc/src/rpc/client/channel.rs
```diff
@@ -4,6 +4,7 @@
 // license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
+use crate::trace::SpanId;
```
tikue: Let's put this in the `use` block on the next line down.

README.md
```diff
@@ -41,13 +41,13 @@ Some other features of tarpc:
...
 
 ### Usage
 Add to your `Cargo.toml` dependencies:
 
 ```toml
-tarpc = "0.18.0"
+tarpc = { version = "0.18.0", features = ["full"] }
```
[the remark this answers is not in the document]
vorot93: We should but only after all other PRs are in

tarpc/Cargo.toml
vorot93: bitcode transport

PR #282, opened 2019-11-28
Improve Travis configuration

vorot93:
This improves Travis configuration by removing deprecated features and adding additional targets.


tikue:
Thanks! Looks like there are some tests that don't pass on Windows.

PR #274, opened 2019-10-10
Unified generic transport

vorot93:
This PR obsoletes the JSON and bincode transports and instead introduces the unified transport that is generic over any `tokio-serde` serialization format as well as `AsyncRead` + `AsyncWrite` medium. This comes with a slight hit for usability (having to manually specify the underlying transport and a pair of encoder and decoder) but it can easily be alleviated by making a custom freestanding connect and listen fns within your own project (which you should do anyway imo).

PR #284, opened 2019-12-04
Update to async-bincode 0.5

iovxw:
We have to include `bincode::Error` since `associated_type_bounds` is unstable.
And `bincode_transport::listen` is an async function now.


vorot93:
Thank you for the pull request!

Upgrade to the latest futures is welcome. However, what benefits does it provide over [shipping a unified transport that is generic over serialization format and the transfer medium](https://github.com/google/tarpc/pull/274)?

iovxw:
Wow! didn't see it, it's awesome

PR #296, opened 2020-03-19
expose some structs to be able to use a non-tokio executor

vhdirk:
For Inox(https://github.com/vhdirk/inox/) I'm combing tarpc with the glib executor as opposed to the default tokio executor.

In order to also have deadlines for requests, I needed to fork ```ClientHandler``` in my own project. That requires some changes in tarpc so I can create, for instance, a response struct. See https://github.com/vhdirk/inox/blob/master/inox-gtk/src/webextension/rpc.rs

I'm still working on the client side. That one is harder since tokio's ```TimeOut```s are not compatible with custom executors, but they are used throughout the (generated) client code.

### DIFF
--- a/tarpc/src/rpc/mod.rs
+++ b/tarpc/src/rpc/mod.rs
@@ -28,7 +28,7 @@ pub mod client;
 pub mod context;
 pub mod server;
 pub mod transport;
-pub(crate) mod util;
+pub mod util;
 
 pub use crate::{client::Client, server::Server, trace, transport::sealed::Transport};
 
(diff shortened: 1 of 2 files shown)


tarpc/src/rpc/mod.rs
```diff
@@ -28,7 +28,7 @@ pub mod client;
 pub mod context;
 pub mod server;
 pub mod transport;
-pub(crate) mod util;
+pub mod util;
```
tikue, line 31: What did you need util for? This module isn't intended to be publicly exposed.

tikue:
Sorry for the long wait! Given that we don't currently have a better way to make tarpc executor agnostic, I think I'm basically okay with this. I'm a little concerned about forward compatibility hazards, but not sufficiently concerned to block this.

tikue:
@vhdirk I'm sorry this languished for so long! :( I think it's best at this point to close it. I think there's still a conversation to be had about how best to support different executors.

Issue #297, opened 2020-03-19
Non-tokio executor & timeouts


vhdirk:
When using an executor that is different from tokio (in my case glib), launching a request from the client results in the following message at runtime:
```
050, marker: <Phantomcow> }), cache: ThreadCache { is_unread: true, has_attachment: false, tags: [] } }))
 INFO  inox_gtk::components::thread_view                 > load_thread: Thread { thread: Some(Thread { ptr: 0x55a29e2a0050, marker: <Phantomcow> }), cache: ThreadCache { is_unread: true, has_attachment: false, tags: [] } }
thread 'main' panicked at 'there is no timer running, must be called from the context of Tokio runtime', /home/dirk/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-0.2.13/src/time/driver/handle.rs:24:9
stack backtrace:```

It seems tokio timeouts are not compatible with other runtimes. It'd be cool if there was a runtime-agnostic solution. Peraps https://github.com/async-rs/futures-timer ?

vorot93:
It is possible to use your own runtime, provided that [it is within the context of Tokio](https://github.com/libp2p/rust-libp2p/issues/1318#issuecomment-564636632).

stjepang:
@vorot93 The problem with the Tokio context is twofold:

1. It is either really difficult or impossible to set up the Tokio context inside your custom executor.
2. There are no practical examples, documentation, or guides on how to do that.

Speaking as a previous Tokio developer and somebody who has been building the async Rust ecosystem for the past several years, I have no idea how to make a "hello world" example of [`futures::executor::ThreadPool`](https://docs.rs/futures/0.3.5/futures/executor/struct.ThreadPool.html) with the Tokio context set up inside it. It's most likely not possible at all!

So until the Tokio documentation shows the most basic "hello world" of how to do that, it is **not** possible to use your own runtime with Tokio.

LucioFranco:
Hey, so I spent a few minutes this morning throwing together an example of using tokio with the futures executor `ThreadPool`. This can be made generic to any executor as long as you wrap the future as shown in the example.

https://gist.github.com/LucioFranco/ec8c6617b7193fd062d1ddc32b5bb991

Most of the complexity here is making it easy to spawn the future with a global runtime. Otherwise, the main thing here is just wrapping the `poll` call in a `handle.enter(|| ...)` like carl showed. 

Hopefully, this is helpful! We could probably add this as an example to the tokio repo.

[...]

PR #298, opened 2020-03-20
use executor-agnostic timers

vhdirk:
This PR replaces the tokio Timeout with a pure futures-based approach: race the request with a delay from the `futures-timer` crate.

I haven't extensively tested this, but for my minimal use case it seems to work.

Supersedes https://github.com/google/tarpc/pull/296 and fixes https://github.com/google/tarpc/issues/297

### DIFF
--- a/tarpc/Cargo.toml
+++ b/tarpc/Cargo.toml
@@ -28,6 +28,7 @@ travis-ci = { repository = "google/tarpc" }
 [dependencies]
 fnv = "1.0"
 futures = "0.3"
+futures-timer = "3.0"
 humantime = "1.0"
 log = "0.4"
 pin-project = "0.4"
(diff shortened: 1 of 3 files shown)


tikue:
Thank you! Things have been a bit crazy here but I'll try to get a review done this weekend.

vorot93:
`futures-timer` tracks timers in its own thread which makes it a mini-runtime on its own. Therefore I am strongly opposed to merging this because this would mean much bigger overhead to placate a minority of users (`tokio` is used by more async prod than all other runtimes combined).

A better solution would be to generalize the timer behaviour, but doing that would be non-trivial.

tarpc/Cargo.toml
```diff
@@ -28,6 +28,7 @@ travis-ci = { repository = "google/tarpc" }
 [dependencies]
 fnv = "1.0"
 futures = "0.3"
+futures-timer = "3.0"
```
LucioFranco: We've seen some issues come from this crate in the past just FYI. I would advise against using it.

vorot93:
Yeah, I don't think that futures-timer is a desirable solution. Thanks for the effort though!

PR #302, opened 2020-05-15
Add tarpc::server proc-macro as syntactic sugar for async methods.

xfbs:
The `tarpc::server` proc-macro can be used to annotate implementations of
services to turn async functions into the proper declarations needed
for tarpc to be able to call them.

### DIFF
--- a/plugins/src/lib.rs
+++ b/plugins/src/lib.rs
@@ -22,8 +22,8 @@ use syn::{
     parse_macro_input, parse_quote, parse_str,
     punctuated::Punctuated,
     token::Comma,
-    Attribute, FnArg, Ident, Lit, LitBool, MetaNameValue, Pat, PatType, ReturnType, Token, Type,
-    Visibility,
+    Attribute, FnArg, Ident, ImplItem, ImplItemMethod, ImplItemType, ItemImpl, Lit, LitBool,
+    MetaNameValue, Pat, PatType, ReturnType, Token, Type, Visibility,
 };
 
 struct Service {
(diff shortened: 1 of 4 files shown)


xfbs:
@googlebot I signed it!

plugins/src/lib.rs
```diff
@@ -212,6 +212,126 @@ pub fn service(attr: TokenStream, input: TokenStream) -> TokenStream {
...
+
+    // generate an identifier consisting of the method name to CamelCase with
+    // Fut appended to it.
+    let fut_name = snake_to_camel(&method.sig.ident.unraw().to_string()) + "Fut";
+    let fut_name_ident = Ident::new(&fut_name, method.sig.ident.span());
+
+    // generate the updated return signature.
+    method.sig.output = parse_quote! {
+        -> ::core::pin::Pin<Box<
+                dyn ::core::future::Future<Output = #ret> + ::core::marker::Send
```
tikue, line 234: Does it need `'static` too?

plugins/src/lib.rs
```diff
@@ -212,6 +212,126 @@ pub fn service(attr: TokenStream, input: TokenStream) -> TokenStream {
...
+///
+/// #[tarpc_plugins::server]
+/// impl World for HelloServer {
+///     async fn hello(self, _: context::Context, name: String) -> String {
+///         format!("Hello, {}! You are connected from {:?}.", name, self.0)
+///     }
+/// }
+/// ```
+///
+/// Into code like this, which is expected by tarpc:
```
tikue, line 278:
```suggestion
/// Into code like this, which matches the service trait definition:
```

tarpc/src/lib.rs
```diff
@@ -233,3 +233,59 @@ pub mod trace;
...
+///
+/// #[tarpc::server]
+/// impl World for HelloServer {
+///     async fn hello(self, _: context::Context, name: String) -> String {
+///         format!("Hello, {}! You are connected from {:?}.", name, self.0)
+///     }
+/// }
+/// ```
+///
+/// Into code like this, which is expected by tarpc:
```
tikue, line 262:
```suggestion
/// Into code like this, which matches the service trait definition:
```

plugins/src/lib.rs
xfbs: I'm not sure, actually. The async-trait crate does a bunch more to transform the function signature and return type, which I don't really (yet) understand.

tikue:
Thank you very much for this great addition. I'm excited to use it!

PR #312, opened 2020-07-24
Add concurrent tests using `join` and `join_all`

joncinque:
These tests are essentially copies of the `concurrent` test,
specifically using `join` and `join_all`.  Note that for the `join_all`
example to work, all of the `Client` clones must be created before *any*
requests are added, otherwise there will be a lifetime problem with the
second request, saying that second client, `c2`, is still borrowed when
`req1` is dropped.  It would require a larger redesign to fix this
issue.

Let me know if you have any questions or suggestions!  I can add a comment to the `join_all` test about the lifetime issue if you like.

### DIFF
--- a/tarpc/tests/service_functional.rs
+++ b/tarpc/tests/service_functional.rs
@@ -1,6 +1,6 @@
 use assert_matches::assert_matches;
 use futures::{
-    future::{ready, Ready},
+    future::{join_all, ready, Ready},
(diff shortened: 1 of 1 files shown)


joncinque:
Subscribing @garious per his request

tikue:
Hey, thanks for the added test coverage!

PR #322, opened 2020-10-23
Add example for custom transport usage

### LINKED ISSUE #234: UNIX sockets
At System76, I develop system services for Linux, and the desktop applications which connect to them. I'd like to move away from using the dbus RPC mechanism on Linux, and am currently weighing my options between varlink and this. However, I need support for UNIX sockets.

These system services are deployed on every installation of the Pop!_OS Linux desktop. We have various services written in Rust, such as the firmware daemon, power daemon, process scheduler daemon, etc. Rather than exposing these services through opened TCP sockets on end users desktops, we have that communication being performed solely through UNIX sockets managed by dbus.

Looking through the existing examples, I don't see a way to set up RPC based on UNIX sockets. Varlink supports passing an address as a string, such as `unix:@com.System76.ServiceName`, yet it seems that tarpc's `Transport` trait requires IP addresses, which doesn't quite map to what I'm doing here. Am I mistaken, and is it possible to use UNIX sockets with tarpc?

l4l:
Took quite some time to do it, hope this can be helpful.

Also probably resolves #234

### DIFF
--- a/tarpc/examples/custom_transport.rs
+++ b/tarpc/examples/custom_transport.rs
@@ -4,7 +4,7 @@ use tarpc::serde_transport as transport;
 use tarpc::server::{BaseChannel, Channel};
 use tokio::net::{UnixListener, UnixStream};
 use tokio_serde::formats::Bincode;
-use tokio_util::codec::{length_delimited::LengthDelimitedCodec, Framed};
+use tokio_util::codec::length_delimited::LengthDelimitedCodec;
 
 #[tarpc::service]
 pub trait PingService {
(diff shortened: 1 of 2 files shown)


l4l:
@googlebot I signed it!

tarpc/examples/custom_transport.rs
```diff
@@ -0,0 +1,60 @@
...
+async fn main() -> std::io::Result<()> {
+    let bind_addr = "/tmp/tarpc_on_unix_example.sock";
+
+    let _ = std::fs::remove_file(bind_addr);
+
+    let (tx_started, rx_started) = tokio::sync::oneshot::channel();
+
+    tokio::spawn(async move {
+        let listener = UnixListener::bind(bind_addr).unwrap();
+        let codec_builder = LengthDelimitedCodec::builder();
```
tikue, line 35: nit: if you do these 2 lines above the spawned async block, then you don't need to signal via channel when the UnixListener is ready.

tikue (approved):
This is excellent, thanks so much! Do you think there is any documentation that could be added to tarpc to make it easier to understand how to do this?

l4l:
> Do you think there is any documentation that could be added to tarpc to make it easier to understand how to do this?

Well, for me it definitely much easier to read and understand an example/tests. I would say it took me quite some time, because of not so obvious generic parameters, e.g. argument of type `Transport<Response<Resp>, ClientMessage<Req>>` looks fairly intuitive, but it's fairly hard to really construct this type. Firstly I tried to do this with `Server::incoming`, but came up with a (cognitive) simpler solution.

PR #327, opened 2020-10-30
Remove `#[derive(Debug)]` from library structs

### LINKED ISSUE #326: Remove the #[derive(Debug)] on library structs
Placing a `#[derive(Debug)]` on user facing library structs that have generics puts an unnecessary burden on users of the library because now the user's type must implement `Debug` even if it has no business doing so.

It's completely fine to `#[derive(Debug)]` on structs that are not generic and/or hold instances of types constructed by users.

Is there any particular reason why the structs in `rpc/server.rs` for example derive Debug?

chansuke:
Fixes #326.

### DIFF
--- a/tarpc/src/rpc.rs
+++ b/tarpc/src/rpc.rs
@@ -4,7 +4,7 @@
 // license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
-#![deny(missing_docs, missing_debug_implementations)]
+#![deny(missing_docs)]
 
 //! An RPC framework providing client and server.
 //!
--- a/tarpc/src/rpc.rs
+++ b/tarpc/src/rpc.rs
@@ -4,7 +4,7 @@
 // license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
-#![deny(missing_docs)]
+#![deny(missing_docs, missing_debug_implementations)]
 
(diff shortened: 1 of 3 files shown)


tarpc/src/rpc.rs
```diff
@@ -4,7 +4,7 @@
 // license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
-#![deny(missing_docs, missing_debug_implementations)]
```
tikue, line 7: Let's leave this in

tarpc/src/rpc/server.rs
```diff
@@ -236,6 +240,12 @@ where
     }
 }
 
+impl<Req, Resp, T> fmt::Debug for BaseChannel<Req, Resp, T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "NewClient {:?}", self)
```
tikue, line 245:
```suggestion
        write!(f, "BaseChannel")
```

tikue (changes requested):

tarpc/src/rpc.rs
chansuke: Ok!

chansuke:
@tikue Thank you for the review.

tikue:
Woo! Thank you!

PR #330, opened 2020-11-04
re-export serde

### LINKED ISSUE #329: tarpc should reexport serde
The tarpc serde derive functionality should not require a a crate to depend on serde. This is possible by overriding the path that the serialize derive codegen uses for the serde crate: https://github.com/serde-rs/serde/pull/1499

Frederik-Baetens:
fixes #329 

Also fixes some issues with old feature flags "serde", it seems like they should be serde1.

### DIFF
--- a/tarpc/Cargo.toml
+++ b/tarpc/Cargo.toml
@@ -15,7 +15,7 @@ description = "An RPC framework for Rust with a focus on ease of use."
 [features]
 default = []
 
-serde1 = ["tarpc-plugins/serde1", "serde", "serde/derive"]
+serde1 = []
 tokio1 = ["tokio/rt-multi-thread"]
 serde-transport = ["tokio-serde", "tokio-util/codec"]
 tcp = ["tokio/net", "tokio/stream"]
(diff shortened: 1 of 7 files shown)


Frederik-Baetens:
@googlebot I signed it!

Frederik-Baetens:
I think the failing test is something that I should fix in the tests, but idk what to change.

tarpc/Cargo.toml
```diff
@@ -33,9 +33,9 @@ humantime = "2.0"
 log = "0.4"
 pin-project = "1.0"
 rand = "0.7"
-serde = { optional = true, version = "1.0", features = ["derive"] }
+serde = { version = "1.0", features = ["derive"] }
 static_assertions = "1.1.0"
-tarpc-plugins = { path = "../plugins", version = "0.8" }
+tarpc-plugins = { path = "../plugins", version = "0.8", features = ["serde1"] }
```
tikue, line 38: I don't think we want to unilaterally enable the serde1 feature. Otherwise, what's the point in having the feature? :) making it optional can decrease compilation time for those users who do not use serde.

tarpc/Cargo.toml
```diff
@@ -33,9 +33,9 @@ humantime = "2.0"
 log = "0.4"
 pin-project = "1.0"
 rand = "0.7"
-serde = { optional = true, version = "1.0", features = ["derive"] }
+serde = { version = "1.0", features = ["derive"] }
```
tikue, line 36: See below comment

tikue (changes requested):

tikue (approved):
Thank you!

PR #332, opened 2020-11-09
Reexport/tokio serde

Frederik-Baetens:


example-service/src/server.rs
```diff
@@ -11,11 +11,11 @@ use std::{
     io,
     net::{IpAddr, SocketAddr},
 };
+use tarpc::tokio_serde::formats::Json;
```
tikue, line 14: nit: put this in the `use tarpc::{ ... }` block below.

example-service/src/client.rs
```diff
@@ -6,8 +6,8 @@
 
 use clap::{App, Arg};
 use std::{io, net::SocketAddr};
+use tarpc::tokio_serde::formats::Json;
```
tikue, line 9: nit: put this in the `use tarpc::{ .. }` block below.

tikue (changes requested):
Two very minor nits, otherwise LGTM :)

example-service/src/client.rs
Frederik-Baetens: Both done :)

tikue:
Thank you!

PR #333, opened 2020-11-11
add serde derivation alias macro

Frederik-Baetens:
I'm not quite sure whether the name i've chosen for the macro is good. But I couldn't immediately come up with a better one.

### DIFF
--- a/plugins/src/lib.rs
+++ b/plugins/src/lib.rs
@@ -216,6 +216,19 @@ impl Parse for DeriveSerde {
 }
 
 /// Generates:
+/// - derive of Debug, serde Serialize & Deserialize
+/// - serde crate annotation
+#[proc_macro_attribute]
+pub fn derive_serde(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let mut gen: proc_macro2::TokenStream = quote! {
+        #[derive(Debug, tarpc::serde::Serialize, tarpc::serde::Deserialize)]
+        #[serde(crate = "tarpc::serde")]
+    };
+    gen.extend(proc_macro2::TokenStream::from(item));
+    proc_macro::TokenStream::from(gen)
+}
+
+/// Generates:
 /// - service trait
 /// - serve fn
 /// - client stub struct
--- a/plugins/src/lib.rs
+++ b/plugins/src/lib.rs
@@ -221,7 +221,7 @@ impl Parse for DeriveSerde {
 #[proc_macro_attribute]
 pub fn derive_serde(_attr: TokenStream, item: TokenStream) -> TokenStream {
     let mut gen: proc_macro2::TokenStream = quote! {
-        #[derive(Debug, tarpc::serde::Serialize, tarpc::serde::Deserialize)]
+        #[derive(tarpc::serde::Serialize, tarpc::serde::Deserialize)]
         #[serde(crate = "tarpc::serde")]
     };
     gen.extend(proc_macro2::TokenStream::from(item));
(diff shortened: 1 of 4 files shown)


plugins/src/lib.rs
```diff
@@ -215,6 +215,19 @@ impl Parse for DeriveSerde {
     }
 }
 
+/// Generates:
+/// - derive of Debug, serde Serialize & Deserialize
+/// - serde crate annotation
+#[proc_macro_attribute]
```
tikue, line 221: I think this should be a `proc_macro_derive` -- https://doc.rust-lang.org/reference/procedural-macros.html#derive-macros

tikue (changes requested):
Thank you for working on this!

plugins/src/lib.rs
Frederik-Baetens: From what I've read derive macros allow you to append something to a strict/enum, so that wouldn't allow me to just substitute the 2 needed attributes.

I might be able to do it with a derive, but I'd think id be copying serdes implementation of that macro then.

I haven't seen how to do macro aliasing with derive macros

tarpc/tests/dataservice.rs
```diff
@@ -0,0 +1,57 @@
+use futures::{future, prelude::*};
+use std::io;
+use tarpc::{
+    client, context,
+    server::{self, Handler},
+};
+
+#[cfg(all(feature = "serde1", feature = "tokio1"))]
```
tikue, line 8: Rather than put the cfgs in the test itself, you can put them in the Cargo.toml. Something like this:

```toml
[[test]]
name = "dataservice"
required-features = ["serde1", "tokio1"]
```

plugins/src/lib.rs
```diff
@@ -215,6 +215,19 @@ impl Parse for DeriveSerde {
...
+pub fn data(_attr: TokenStream, item: TokenStream) -> TokenStream {
```
tikue, line 222: let's just call it `derive_serde` to be explicit about what it's doing.

tikue (changes requested):

plugins/src/lib.rs
Frederik-Baetens: Done :)

tarpc/tests/dataservice.rs
Frederik-Baetens: Done :)

tarpc/tests/dataservice.rs
```diff
@@ -0,0 +1,52 @@
...
+        match color {
+            TestData::White => TestData::Black,
+            TestData::Black => TestData::White,
+        }
+    }
+}
+
+#[tokio::test]
+async fn test_call() -> io::Result<()> {
+    let (client_transport, server_transport) = tarpc::transport::channel::unbounded();
```
tikue, line 35: This channel is an in-memory channel and does't perform serialization, so I don't think it's actually testing that the serde derive works correctly.

Frederik-Baetens: Oh, i'll change that to a serde based channel then :)

plugins/src/lib.rs
```diff
@@ -215,6 +215,19 @@ impl Parse for DeriveSerde {
     }
 }
 
+/// Generates:
+/// - derive of Debug, serde Serialize & Deserialize
+/// - serde crate annotation
+#[proc_macro_attribute]
+pub fn derive_serde(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let mut gen: proc_macro2::TokenStream = quote! {
+        #[derive(Debug, tarpc::serde::Serialize, tarpc::serde::Deserialize)]
```
tikue, line 224: tiny nit: I don't think there's any reason to derive Debug here rather than allowing users to derive Debug themselves. The main reason we reexport the serde derives is to avoid versioning conflicts.

tikue (approved):

PR #340, opened 2021-01-20
mipsel

### LINKED ISSUE #339: AtomicU64 availability
on MIPS, `AtomicU64` is not available and thus, this crate does not compile.
I could make a PR to fix that but first I need to know which solution you prefer:
1. switch to `AtomicUsize`
2. use `#[cfg(target_arch)]` to use `AtomicU32` for MIPS
3. use a 64bit Mutex
4. use a Generic

Also, in my quick n' dirty solution where I did `1.` I then casted the resulting `usize` to `u64` so I don't have to change the whole API. How do we go about that?

M1cha:
fixes #339


M1cha:
@googlebot I signed it!

tarpc/src/rpc/client/channel.rs
```diff
@@ -24,11 +24,20 @@ use std::{
...
+        atomic::{AtomicUsize, Ordering},
         Arc,
     },
 };
 
+#[allow(dead_code)]
+#[allow(clippy::no_effect)]
+const CHECK_USIZE: bool = {
+    if std::mem::size_of::<usize>() > std::mem::size_of::<u64>() {
+        ["usize is big small to fit in u64"][42];
```
tikue, line 36:
```suggestion
        ["usize is too big to fit in u64"][42];
```

tikue, line 36: Nit: consider using the [compile_error](https://doc.rust-lang.org/std/macro.compile_error.html) macro.

tikue (changes requested):
Great, thanks so much! One minor nit

tikue:
Also, do you want to add MIPS as an architecture that's tested in CI/CD? Otherwise this support could be inadvertently broken in the future.

tarpc/src/rpc/client/channel.rs
tikue: ah, I see. Ok, this works then! Perhaps add a TODO?

M1cha:
FYI: the clippy fail is caused by code that I did not change. I don't know why didn't come up before.

I'll add a MIPS check to the workflow, good idea

M1cha:
Do you want me to fix the clippy warning?
If yes, should this be part of this or a separate PR?

tikue:
Not sure why Clippy is complaining now! Yes, feel free to fix it, since it's just a one line change. Thanks!

tarpc/src/rpc/client/channel.rs
```diff
@@ -106,7 +115,7 @@ impl<Req, Resp> Channel<Req, Resp> {
 
         let (response_completion, response) = oneshot::channel();
         let cancellation = self.cancellation.clone();
-        let request_id = self.next_request_id.fetch_add(1, Ordering::Relaxed);
+        let request_id = self.next_request_id.fetch_add(1, Ordering::Relaxed) as u64;
```
tikue, line 118:
```suggestion
        let request_id = u64.try_from(self.next_request_id.fetch_add(1, Ordering::Relaxed)).unwrap();
```

Even though there's already a compile-time check, in the spirit of defense in depth, I'd like to explicitly check that the conversion is sound.

tikue (changes requested):
Sorry, just one last thing!

tarpc/src/rpc/client/channel.rs
M1cha: Additionally, looking at the implementation of try_from for targets with usize <= 64, they return `X as u64` unconditionally.
So in theory compiler optimizations should get rid of both try_from and unwrap anyway which means the code would be just as fast as before :)

M1cha: for reference: https://github.com/rust-lang/rust/blob/1.49.0/library/core/src/convert/num.rs

tikue (approved):

tikue:
Hey @M1cha, heads up that mipsel was [downgraded to tier 3](https://github.com/rust-lang/compiler-team/issues/648), which has caused [check failures on pull requests](https://github.com/google/tarpc/pull/407), so I plan to remove the workflow that checks that the project compiles on mipsel. Sorry about that!

PR #343, opened 2021-03-17
Reduce required tokio features

l4l:
Current example fails to build with minimal set of required tokio features, i.e. `features = ["macros", "net", "rt-multi-thread"]` because tokio tests are not behind a `cfg(test)` gate. For some reason this issue appeared for me after I tried to upgrade tarpc 0.24.1 -> 0.25 dependency. The pr so far consists of 3 commits:

1. All of the `tokio::test`-s are moved inside a `#[cfg(test) mod test { .. }`. This commit is kinda big, maybe it's better to make it smaller with adding a cfg for each test instead.
2. Reduces tokio features for example-service.
3. This one removes a version part of tarpc dependency of example-service, otherwise it fetches from crates-io. I'm not sure how it worked before, perhaps this one is not really needed.

### DIFF
--- a/tarpc/src/server/throttle.rs
+++ b/tarpc/src/server/throttle.rs
@@ -171,177 +171,179 @@ where
 }
 
 #[cfg(test)]
-use super::testing::{self, FakeChannel, PollExt};
-#[cfg(test)]
-use crate::Request;
-#[cfg(test)]
-use pin_utils::pin_mut;
-#[cfg(test)]
-use std::{marker::PhantomData, time::Duration};
+mod tests {
+    use super::*;
 
-#[tokio::test]
-async fn throttler_in_flight_requests() {
-    let throttler = Throttler {
-        max_in_flight_requests: 0,
-        inner: FakeChannel::default::<isize, isize>(),
-    };
+    use super::super::in_flight_requests::AlreadyExistsError;
+    use super::super::testing::{self, FakeChannel, PollExt};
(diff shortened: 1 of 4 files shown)


tikue:
Thank you! From a quick glance this looks good. I'll review more in depth later today.

tarpc/src/server/throttle.rs
```diff
@@ -171,177 +171,179 @@ where
...
-    for i in 0..5 {
-        throttler
-            .inner
-            .in_flight_requests
-            .start_request(i, SystemTime::now() + Duration::from_secs(1))
-            .unwrap();
+mod tests {
+    use super::*;
+
+    use super::super::in_flight_requests::AlreadyExistsError;
```
tikue, line 177: IMO, it's simpler to do `crate::client::`

tarpc/src/server/throttle.rs
```diff
@@ -171,177 +171,179 @@ where
...
+    use super::super::testing::{self, FakeChannel, PollExt};
```
tikue, line 178: can you combine all `use crate::...` lines into a single use block?

tikue (changes requested):
Looks good! I was able to check the diff locally, ignoring whitespace, and it was much easier to see that little has changed.

tikue:
I'm going to merge as-is and fix the nits in a followup commit.

PR #344, opened 2021-03-25
Replace badly supported u128 TraceId with byte array

Kouzukii:
Binary serde codecs such as [CBOR](https://github.com/pyfisch/cbor) do not support Rusts u128/i128 datatypes (https://github.com/pyfisch/cbor/issues/140). 
In order to be able to use these with tarpc, the u128 TraceId could be made serializable using a 16 byte array as the internal representation.

### DIFF
--- a/tarpc/src/trace.rs
+++ b/tarpc/src/trace.rs
@@ -48,7 +48,7 @@ pub struct Context {
 /// same trace ID.
 #[derive(Debug, Default, PartialEq, Eq, Hash, Clone, Copy)]
 #[cfg_attr(feature = "serde1", derive(serde::Serialize, serde::Deserialize))]
-pub struct TraceId(u128);
+pub struct TraceId(#[cfg_attr(feature = "serde1", serde(with = "u128_serde"))] u128);
 
 /// A 64-bit identifier of a span within a trace. The identifier is unique within the span's trace.
 #[derive(Debug, Default, PartialEq, Eq, Hash, Clone, Copy)]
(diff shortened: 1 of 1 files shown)


tikue:
Thanks for the pr! What about keeping the u128 but having custom (de)serialize fns using the `#[serde(with)]` annotation?

tarpc/src/trace.rs
```diff
@@ -47,7 +47,10 @@ pub struct Context {
 /// A 128-bit UUID identifying a trace. All spans caused by the same originating span share the
 /// same trace ID.
 #[derive(Debug, Default, PartialEq, Eq, Hash, Clone, Copy)]
-#[cfg_attr(feature = "serde1", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(
```
tikue, line 50: I'd prefer to just have one serde impl instead of adding a feature. I'm ok making a breaking change.

tarpc/src/trace.rs
```diff
@@ -95,3 +98,46 @@ impl fmt::Display for SpanId {
...
+                let bytes: [u8; 16] = seq.next_element()?.ok_or_else(|| {
+                    serde::de::Error::invalid_length(0, &"tuple struct TraceId with 1 element")
+                })?;
+                Ok(TraceId(u128::from_le_bytes(bytes)))
+            }
+        }
+
+        deserializer.deserialize_tuple_struct("TraceId", 1, Visitor)
+    }
+}
```
tikue, line 143: I think this could be made simpler by using the `#[serde(with)]` attribute on the u128 field. e.g.

```rust
pub struct TraceId(#[cfg_attr(feature = "serde1", serde(with = "u128_serde"))] u128);
```

and then we'd just need a `mod u128_serde` with the serialize and deserialize fns:

```suggestion
#[cfg(feature = "serde1")]
mod u128_serde {
    pub fn serialize<S>(u: &u128, serializer: S) -> Result<S::Ok, S::Error>
        where S: serde::Serializer
    {
        serde::Serialize::serialize(&u.to_le_bytes(), serializer)
    }

    pub fn deserialize<'de, D>(deserializer: D) -> Result<u128, D::Error>
        where D: serde::Deserializer<'de>
    {
        Ok(u128::from_le_bytes(serde::Deserialize::deserialize(deserializer)?))
    }
}

```

tikue (changes requested):

tarpc/src/trace.rs
Kouzukii: that is indeed much simpler, I was too focused on keeping it the same as the expanded Serialize derive. If you don't mind the breaking change I'll update my PR.

tikue: Ah, that makes sense—thanks for the context! I appreciate you striving to keep the implementation the same.

My preference is to keep things simple. Thanks for the updates!

tikue (approved):

PR #345, opened 2021-03-27
Do not enable tokio-serde/json by default

oblique:
By having `tokio-serde/json` forces users that want to use `bincode` only, to have extra dependencies.

After this commit, users need to specify the format they are interested in their Cargo.toml.


oblique:
@googlebot I signed it!

tikue:
Thanks for the pr! I agree it should be opt in.

PR #346, opened 2021-05-06
add serde-transport-json feature flag

Frederik-Baetens:
In general, it should be possible to use, or at least import all functionality of a library, when having only that library in your cargo.toml

This means that commits which add things to the Cargo.toml, without also adding corresponding usage of those newly added libraries are usually a red flag.

This fixes such a problem introduced in #345, where usage of the serde_json feature tokio_serde::formats::Json required an additional import of tokio-serde with the json feature flag, rather than setting an appropriate tarpc feature flag.

This still doesn't enable the json featureflag by default, so the creator of #345 should still be happy


tikue:
Thanks :)

PR #353, opened 2021-11-08
Re-add typo fixes

scott-kirk:
The typo fixes that were added by commit b5d9aaa
were accidentally reverted by commit 1e680e3, this
will add them back


tikue:
Ah, thank you!

PR #358, opened 2021-11-29
Fix a compiling issue of the official example

shi-yan:
because of the following error :

```bash
error[E0599]: the method `execute` exists for struct `BaseChannel<_, _, UnboundedChannel<ClientMessage<_>, Response<_>>>`, but its trait bounds were not satisfied
  --> src/main.rs:39:25
   |
39 |     tokio::spawn(server.execute(HelloServer.serve()));
   |                         ^^^^^^^ method cannot be called on `BaseChannel<_, _, UnboundedChannel<ClientMessage<_>, Response<_>>>` due to unsatisfied trait bounds
   |
   = note: the following trait bounds were not satisfied:
           `<&BaseChannel<_, _, UnboundedChannel<ClientMessage<_>, Response<_>>> as futures::Stream>::Item = _`
           which is required by `&BaseChannel<_, _, UnboundedChannel<ClientMessage<_>, Response<_>>>: tarpc::server::incoming::Incoming<_>`
           `&BaseChannel<_, _, UnboundedChannel<ClientMessage<_>, Response<_>>>: futures::Stream`
           which is required by `&BaseChannel<_, _, UnboundedChannel<ClientMessage<_>, Response<_>>>: tarpc::server::incoming::Incoming<_>`
   = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
   |
1  | use tarpc::server::Channel;
   |

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ rather than in a separate language such as .proto. This means there's no separat
 process, and no context switching between different languages.
 
 Some other features of tarpc:
-- Pluggable transport: any type impling `Stream<Item = Request> + Sink<Response>` can be
+- Pluggable transport: any type implementing `Stream<Item = Request> + Sink<Response>` can be
   used as a transport to connect the client and server.
 - `Send + 'static` optional: if the transport doesn't require it, neither does tarpc!
 - Cascading cancellation: dropping a request will send a cancellation message to the server.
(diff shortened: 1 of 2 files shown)


tikue:
Ah, thank you! Actually, the README is generated from the doc comment in tarpc/src/lib.rs — which also makes it clearer what happened here: Channel was imported on line 153:

https://github.com/google/tarpc/blob/f7c5d6a7c378565c7f0123dc2b9b5960b15d38bf/tarpc/src/lib.rs#L151-L154

because those are the imports for the first code snippet that actually calls `execute`. However, as you can see, those imports are "commented out" of the readme, because the prior code snippet (which doesn't actually use Channel and forgot to import it) had already set up the imports on line 89:

https://github.com/google/tarpc/blob/f7c5d6a7c378565c7f0123dc2b9b5960b15d38bf/tarpc/src/lib.rs#L89-L92

So the solution is to add the import to line 89 and then run `cargo sync-readme` from the tarpc crate (tarpc/tarpc).

shi-yan:
Thank you so much for the explanation!

shi-yan:
I have updated the readme by running sync-readme as you suggested, but it seems to have introduced more typos. I can fix those .

tikue:
Awesome, thanks! Yeah I think I merged some prior commits without thinking about this :)

PR #359, opened 2021-12-23
adds PartialEq, Eq and Hash traits to RpcError

dvdsk:
Tiny commit to derive these traits for `RpcError`. The underlying `ServerError` type already derives these traits therefore it should not be a problem to do so for RpcError.

### DIFF
--- a/tarpc/src/client.rs
+++ b/tarpc/src/client.rs
@@ -168,7 +168,7 @@ struct ResponseGuard<'a, Resp> {
 
 /// An error that can occur in the processing of an RPC. This is not request-specific errors but
 /// rather cross-cutting errors that can always occur.
-#[derive(thiserror::Error, Debug, PartialEq, Eq, Hash)]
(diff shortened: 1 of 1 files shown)


dvdsk:
Should `RpcError` also (conditionally) derive `Serialize` and `Deserialize`?. The underlying `ServerError` type does so. I think the compile time impact would be negligible as the serde derive macro will already be compiled anyway.

tikue:
Thanks! I don't see any harm in impling Serialize and Deserialize as well, if you want to add that 👍

dvdsk:
hey! I needed those for a personal project however I can do without them now. 

I do not mind losing the imp however other users may by now be relying on them. The API guidelines also recommend having them (https://rust-lang.github.io/api-guidelines/interoperability.html).

If adding error variants becomes tricky due to longer compile times putting the impl  behind a feature might be an option. I wouldn't mind taking a peek at issue bit.

PR #360, opened 2022-01-06
Use single-threaded Tokio runtime

zakcutner:
As far as I can tell, the multi-threaded runtime is not used anywhere (apart from tests) but sorry if I'm missing something!


tikue:
Thanks! I think you are correct :)

PR #367, opened 2022-05-26
Serialize RPC deadline as a Duration.

### LINKED ISSUE #366: RPC deadline and clock skew on different machines
I have a development machine machine and a Raspberry pi that happens to have its system clock out of sync. All my RPCs from Pi to my machine would fail with `DeadlineExceeded`, because Pi sets the deadline at 2021.

Does it make sense to use relative time (10 seconds) instead of absolute system time to represent RPC deadline? Given typical clock skew between machines could be O(seconds), using absolute time seems quite inaccurate.

tikue:
Duration was previously serialized as `SystemTime`. However, absolute times run into problems with clock skew: if the remote machine's clock is too far in the future, the RPC deadline will be exceeded before request processing can begin. Conversely, if the remote machine's clock is too far in the past, the RPC deadline will not be enforced.

By converting the absolute deadline to a relative duration, clock skew is no longer relevant, as the remote machine will convert the deadline into a time relative to its own clock. This mirrors how the gRPC HTTP2 protocol includes a `Timeout` in the [request headers](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) but the SDK uses [timestamps](https://grpc.io/blog/deadlines/#setting-a-deadline). Keeping the absolute time in the core APIs maintains all the benefits of today, namely, natural deadline propagation between RPC hops when using the current context.

This serialization strategy means that, generally, the remote machine's deadline will be slightly in the future compared to the local machine's deadline. Depending on network transfer latencies, this could be microseconds to milliseconds, or worse. Because the deadline is not intended for high-precision scenarios, I don't view this is as problematic.

Because this change only affects the serialization layer, local transports that bypass serialization are not affected.

Jaeger screenshot showing change in deadline between client and server:

image

Fixes #366

### DIFF
--- a/tarpc/examples/tracing.rs
+++ b/tarpc/examples/tracing.rs
@@ -6,7 +6,6 @@
 
 use crate::{add::Add as AddService, double::Double as DoubleService};
 use futures::{future, prelude::*};
-use std::env;
 use tarpc::{
     client, context,
     server::{incoming::Incoming, BaseChannel},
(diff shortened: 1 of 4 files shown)


ditsing:
Using the serialization layer to hide the conversion is pretty neat. Thanks for fixing this!

tikue:
Thanks for raising this issue and providing prompt feedback, @ditsing !

PR #371, opened 2022-06-25
feat: re-export used packages

### LINKED ISSUE #370: Update outdated dependencies or re-export
It would be great to have the dependency updated like tokio and others. Also, it might be easier on maintainer and users to just re-export.

Right now and maintaining a specific version for the following to ensure compatibility with tarpc

 
```rust
pub use tokio_serde::formats::Json;
pub use tokio_util::codec::length_delimited::LengthDelimitedCodec;
```

I'm happy to open a PR

Thanks

kkharji:
### Problem

Library users might get stuck with or ran into issues while using tarpc because of incompatible third party libraries. in particular, `tokio_serde` and `tokio_util`.

### Solution

This PR hope to achive the following:

1. re-export struct from third party libraries that the end user would most likely import.
2. hide few types behind feature flags like tokio_serde::frmats::Json
3. Update third library packages to latest release and fix resulting issues from that.

### Important Notes

- tokio_util 7.3 [DelayQueue::poll_expired](https://docs.rs/tokio-util/latest/tokio_util/time/delay_queue/struct.DelayQueue.html#method.poll_expired) therefore, `InFlightRequests::poll_expired` now returns `Poll<Option<u64>>`

### Todo 

- [x] update documentation/example with re-export packages
- [ ] bump version

closes #370

### DIFF
--- a/tarpc/src/lib.rs
+++ b/tarpc/src/lib.rs
@@ -209,7 +209,7 @@
 pub use serde;
 
 #[cfg(feature = "serde-transport")]
-pub use tokio_serde;
+pub use {tokio_serde, tokio_util};
 
 #[cfg(feature = "serde-transport")]
 #[cfg_attr(docsrs, doc(cfg(feature = "serde-transport")))]
--- a/tarpc/src/server.rs
+++ b/tarpc/src/server.rs
@@ -393,11 +393,7 @@ where
                 Poll::Pending | Poll::Ready(None) => Closed,
             };
 
-            let expiration_status = match self
-                .in_flight_requests_mut()
-                .poll_expired(cx)
-                .map_err(ChannelError::Timer)?
-            {
+            let expiration_status = match self.in_flight_requests_mut().poll_expired(cx) {
                 // No need to send a response, since the client wouldn't be waiting for one
                 // anymore.
                 Poll::Ready(Some(_)) => Ready,
--- a/example-service/Cargo.toml
+++ b/example-service/Cargo.toml
@@ -18,14 +18,14 @@ anyhow = "1.0"
 clap = { version = "3.0.0-rc.9", features = ["derive"] }
 log = "0.4"
 futures = "0.3"
-opentelemetry = { version = "0.16", features = ["rt-tokio"] }
-opentelemetry-jaeger = { version = "0.15", features = ["rt-tokio"] }
+opentelemetry = { version = "0.17", features = ["rt-tokio"] }
+opentelemetry-jaeger = { version = "0.16", features = ["rt-tokio"] }
 rand = "0.8"
 tarpc = { version = "0.29", path = "../tarpc", features = ["full"] }
 tokio = { version = "1", features = ["macros", "net", "rt-multi-thread"] }
 tracing = { version = "0.1" }
-tracing-opentelemetry = "0.15"
-tracing-subscriber = "0.2"
+tracing-opentelemetry = "0.17"
+tracing-subscriber = "0.3"
 
 [lib]
 name = "service"
--- a/example-service/Cargo.toml
+++ b/example-service/Cargo.toml
@@ -25,7 +25,7 @@ tarpc = { version = "0.29", path = "../tarpc", features = ["full"] }
 tokio = { version = "1", features = ["macros", "net", "rt-multi-thread"] }
 tracing = { version = "0.1" }
 tracing-opentelemetry = "0.17"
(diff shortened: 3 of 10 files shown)


tarpc/src/server.rs
```diff
@@ -393,11 +394,7 @@ where
                 Poll::Pending | Poll::Ready(None) => Closed,
             };
 
-            let expiration_status = match self
-                .in_flight_requests_mut()
-                .poll_expired(cx)
-                .map_err(ChannelError::Timer)?
```
tikue, line 399: Can `ChannelError::Timer` variant be removed from the enum now?

tikue (changes requested):
Thanks for the PR! I'm traveling right now and reviewing on my phone, so please forgive any silly questions!

tarpc/src/lib.rs
```diff
@@ -209,7 +209,7 @@
 pub use serde;
 
 #[cfg(feature = "serde-transport")]
-pub use tokio_serde;
+pub use tokio_serde::formats;
```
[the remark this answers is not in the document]
tami5: hey @tikue, I  believe it's much cleaner and straightforward, I'd go with just exporting the struct but due the the other formats it made sense to me to export module format. Idk, do you think I should just keep it as exporting tokio-serde instead

tarpc/src/lib.rs
```diff
@@ -220,6 +220,12 @@ pub mod trace;
 #[cfg(feature = "serde1")]
 pub use tarpc_plugins::derive_serde;
 
+#[cfg(feature = "serde-transport")]
+pub use tokio_util::codec;
+
+pub use futures::future;
+pub use futures::prelude as future_prelude;
```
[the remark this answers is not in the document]
tami5: Oh indeed, only internal used.  I saw it used in examples, and thought to myself it is something users need

tarpc/src/lib.rs
```diff
@@ -220,6 +220,12 @@ pub mod trace;
 #[cfg(feature = "serde1")]
 pub use tarpc_plugins::derive_serde;
 
+#[cfg(feature = "serde-transport")]
+pub use tokio_util::codec;
```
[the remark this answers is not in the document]
tami5: Yah see why you might prefer that, same with tokio-serde. I will just export them both instead of submodules

tarpc/src/server.rs
```diff
@@ -34,6 +34,7 @@ pub mod limits;
 /// Provides helper methods for streams of Channels.
 pub mod incoming;
 
+// TODO: shouldn't we export this from tokio at top level?
```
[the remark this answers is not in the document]
tami5: Oh my bad through it's tokio module. removing this todo ..

tarpc/src/lib.rs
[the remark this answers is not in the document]
tikue: Personally I would rather re-export the whole crate, so that if the crate's module structure changes we don't have to keep it updated here.

tarpc/src/server.rs
tikue: I mean in the enum definition [here](https://github.com/google/tarpc/blob/8f9ce0d9b698264d6beeb8d1db62947f7818278f/tarpc/src/server.rs#L348). I think `ChannelError::Timer` might no longer be used anywhere.

tikue (approved):
Looks great, thanks!

tikue:
Oh, looks like example-service needs to be updated for these changes, as well: https://github.com/google/tarpc/runs/7168854729?check_suite_focus=true

tarpc/src/server.rs
kkharji: Oh yah, I agree, it should be removed.

kkharji:
> Oh, looks like example-service needs to be updated for these changes, as well: https://github.com/google/tarpc/runs/7168854729?check_suite_focus=true

I will fixed in a bit, don't you hate when API changes 😓.

tarpc/src/server.rs
kkharji: I tried to remove it but felt it might be needed at some point, what do u think

tikue: sure, though it can always be added back :)

kkharji: @tikue would like me to still remove still?

tikue:
Thank you for your contribution and patience with my review!

PR #369, opened 2022-06-23
feat: support unix sockets

kkharji:

PR #380, opened 2022-09-21
feat: Unix domain sockets with serde transports

kbknapp:
Picks up where #369 apparently stalled. Funny fact I didn't look at existing PRs when doing this, so we just happened to pick almost identical impls since it's basically just copying the `tcp` module.

All tests are passing locally, I've also run this through a local application I have using unix sockets and it appears to work as desired.

The unix sockets are behind the feature flag `unix` and in the `unix` module to keep modules similar to those in Tokio and std. They are also gated by the target OS being a unix flavor as well.

~There's one minor additional tag along commit of a UI test that was failing on master.~ (edit: looks like that may have been an artifact of me testing with nightly. It's been removed).

### DIFF
--- a/tarpc/src/serde_transport.rs
+++ b/tarpc/src/serde_transport.rs
@@ -433,22 +433,27 @@ pub mod unix {
     pub struct TempSock(std::path::PathBuf);
 
     impl TempSock {
-        /// A named socket that results in `<name>.sock`
+        /// A named socket that results in `<tempdir>/<name>.sock`
         pub fn new<S: AsRef<str>>(name: S) -> Self {
+            Self::exact(format!("{}.sock", name.as_ref()))
+        }
+
+        /// A named socket that uses the name exactly as passed without allocating a new `String`
+        /// to add `.sock`, this results in `<tempdir>/<name>`
+        pub fn exact<S: AsRef<str>>(name: S) -> Self {
             let mut sock = std::env::temp_dir();
-            sock.push(format!("{}.sock", name.as_ref()));
+            sock.push(name.as_ref());
             Self(sock)
         }
 
-        /// Appends a random hex string to the socket name resulting in `<name>_<xxxxx>.sock`
+        /// Appends a random hex string to the socket name resulting in
+        /// `<tempdir>/<name>_<xxxxx>.sock`
         pub fn with_random<S: AsRef<str>>(name: S) -> Self {
-            let mut sock = std::env::temp_dir();
-            sock.push(format!(
+            Self::exact(format!(
                 "{}_{:x}.sock",
                 name.as_ref(),
                 rand::random::<u64>()
-            ));
-            Self(sock)
+            ))
         }
     }
 
--- a/tarpc/src/serde_transport.rs
+++ b/tarpc/src/serde_transport.rs
@@ -277,6 +277,168 @@ pub mod tcp {
     }
 }
 
+#[cfg(all(unix, feature = "unix"))]
+#[cfg_attr(docsrs, doc(cfg(all(unix, feature = "unix"))))]
+/// Unix Domain Socket support for generic transport using Tokio.
+pub mod unix {
+    use {
+        super::*,
+        futures::ready,
+        std::{marker::PhantomData, path::Path},
+        tokio::net::{unix::SocketAddr, UnixListener, UnixStream},
+        tokio_util::codec::length_delimited,
+    };
+
+    mod private {
+        use super::*;
+
+        pub trait Sealed {}
+
+        impl<Item, SinkItem, Codec> Sealed for Transport<UnixStream, Item, SinkItem, Codec> {}
+    }
+
+    impl<Item, SinkItem, Codec> Transport<UnixStream, Item, SinkItem, Codec> {
+        /// Returns the socket address of the remote half of the underlying [`UnixStream`].
+        pub fn peer_addr(&self) -> io::Result<SocketAddr> {
+            self.inner.get_ref().get_ref().peer_addr()
+        }
+        /// Returns the socket address of the local half of the underlying [`UnixStream`].
+        pub fn local_addr(&self) -> io::Result<SocketAddr> {
+            self.inner.get_ref().get_ref().local_addr()
+        }
+    }
+
+    /// A connection Future that also exposes the length-delimited framing config.
+    #[must_use]
+    #[pin_project]
+    pub struct Connect<T, Item, SinkItem, CodecFn> {
+        #[pin]
+        inner: T,
+        codec_fn: CodecFn,
+        config: length_delimited::Builder,
+        ghost: PhantomData<(fn(SinkItem), fn() -> Item)>,
+    }
+
+    impl<T, Item, SinkItem, Codec, CodecFn> Future for Connect<T, Item, SinkItem, CodecFn>
+    where
+        T: Future<Output = io::Result<UnixStream>>,
+        Item: for<'de> Deserialize<'de>,
+        SinkItem: Serialize,
+        Codec: Serializer<SinkItem> + Deserializer<Item>,
+        CodecFn: Fn() -> Codec,
+    {
+        type Output = io::Result<Transport<UnixStream, Item, SinkItem, Codec>>;
+
+        fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+            let io = ready!(self.as_mut().project().inner.poll(cx))?;
+            Poll::Ready(Ok(new(self.config.new_framed(io), (self.codec_fn)())))
+        }
+    }
+
+    impl<T, Item, SinkItem, CodecFn> Connect<T, Item, SinkItem, CodecFn> {
+        /// Returns an immutable reference to the length-delimited codec's config.
+        pub fn config(&self) -> &length_delimited::Builder {
+            &self.config
+        }
+
+        /// Returns a mutable reference to the length-delimited codec's config.
+        pub fn config_mut(&mut self) -> &mut length_delimited::Builder {
+            &mut self.config
+        }
+    }
+
+    /// Connects to socket named by `path`, wrapping the connection in a Unix Domain Socket
+    /// transport.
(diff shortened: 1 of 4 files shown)


tikue:
Thanks so much! I'm away this week, will hopefully review next week

tarpc/tests/service_functional.rs
```diff
@@ -136,6 +136,45 @@ async fn serde() -> anyhow::Result<()> {
...
 }
 
+#[cfg(all(feature = "serde-transport", feature = "unix", unix))]
+#[tokio::test]
+async fn serde_uds() -> anyhow::Result<()> {
+    use std::fs;
+    use tarpc::serde_transport;
+    use tokio_serde::formats::Json;
+
+    const SOCK: &str = "tarpc_serde.sock";
```
tikue, line 146: What about using a [temp directory](https://doc.rust-lang.org/stable/std/env/fn.temp_dir.html#) as well, to ensure cleanup in case the test crashes in the middle?

```suggestion
    let mut sock = std::env::temp_dir();
    sock.push("tarpc_serde.sock");
```

Perhaps also append some random number to prevent collisions on the off chance that tests are running concurrently?

tarpc/tests/service_functional.rs
```diff
@@ -136,6 +136,45 @@ async fn serde() -> anyhow::Result<()> {
...
+
+    let transport = serde_transport::unix::connect(SOCK, Json::default).await?;
+    let client = ServiceClient::new(client::Config::default(), transport).spawn();
+
+    // Save results using socket so we can clean the socket even if our test assertions fail
+    let res1 = client.add(context::current(), 1, 2).await;
+    let res2 = client.hey(context::current(), "Tim".to_string()).await;
+
+    // Clean temporary socket even if assertions fail
+    let _ = fs::remove_file(SOCK);
```
tikue, line 170: optional nit: Another way to accomplish this would be with a cleanup struct initialized up above, when the socket path is initialized. That way, the cleanup code is a bit out of the way of the main test code.

```suggestion
    // Clean temporary socket even if assertions fail
   struct Cleanup(std::path::PathBuf);
   impl Drop for Cleanup {
       fn drop(&mut self) {
           let _ = fs::remove_file(self.0);
       }
   }
```

tikue (changes requested):
Thanks for your patience! Overall looks good. I left comments about a few minor things...

tarpc/src/serde_transport.rs
```diff
@@ -277,6 +277,168 @@ pub mod tcp {
...
+        tokio_util::codec::length_delimited,
+    };
+
+    mod private {
+        use super::*;
+
+        pub trait Sealed {}
+
+        impl<Item, SinkItem, Codec> Sealed for Transport<UnixStream, Item, SinkItem, Codec> {}
+    }
```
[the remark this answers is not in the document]
kbknapp: Haha funny you ask, because I asked myself the same. Implementing `Sealed` (a private trait) is usually so you can make that a bound on some other trait so no external crates can implement your (non-private) trait on their types. e.g. https://rust-lang.github.io/api-guidelines/future-proofing.html

However here it doesn't look like it's actually a bound anywhere?

I left it in because I didn't know if it was planned for the future, or there was some other reason.

tarpc/tests/compile_fail/serde_transport/must_use_tcp_connect.stderr
```diff
@@ -1,4 +1,4 @@
-error: unused `Connect` that must be used
+error: unused `tarpc::serde_transport::tcp::Connect` that must be used
```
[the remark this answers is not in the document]
kbknapp: My thinking is because prior to this PR there was only one `Connect` type (`tarpc::serde_transport::tcp::Connect`) but now there are two (`tarpc::serde_transport::unix::Connect`, so Rust is giving the full path to disambiguate?

tarpc/src/serde_transport.rs
```diff
@@ -277,6 +277,168 @@ pub mod tcp {
...
+        Item: for<'de> Deserialize<'de>,
+        SinkItem: Serialize,
+        Codec: Serializer<SinkItem> + Deserializer<Item>,
+        CodecFn: Fn() -> Codec,
+    {
+        type Item = io::Result<Transport<UnixStream, Item, SinkItem, Codec>>;
+
+        fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+            let conn: UnixStream =
+                ready!(Pin::new(&mut self.as_mut().project().listener).poll_accept(cx)?).0;
```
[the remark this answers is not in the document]
kbknapp: It does not. TCP also doesn't need it, but should I leave that for different PR to keep these commits related?

tarpc/tests/service_functional.rs
kbknapp: Because the unix sockets have some other requirements like implementing `AsRef<Path>` I went ahead and added a `TempSock` to the `unix` module. If you'd prefer, I can remove that and just duplicate the code in both tests, but it seems like a useful thing to me even for non-test code.

`TempSock`: [0d6fbdb](https://github.com/google/tarpc/pull/380/commits/0d6fbdbec2f8e9213138ce4a26bb132e7e9a2a3e)
Test cleanup: [ce5715c](https://github.com/google/tarpc/pull/380/commits/ce5715c93f0e58eef45bf0f2448b3a0d7f5fda11)

tarpc/tests/service_functional.rs
kbknapp: See [this comment in another thread](https://github.com/google/tarpc/pull/380#discussion_r984657583).

tarpc/src/serde_transport.rs
```diff
@@ -277,6 +277,168 @@ pub mod tcp {
...
+            Poll::Ready(Some(Ok(new(
+                self.config.new_framed(conn),
+                (self.codec_fn)(),
+            ))))
+        }
+    }
+}
+
 #[cfg(test)]
 mod tests {
```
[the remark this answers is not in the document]
kbknapp: Added in [03f7236](https://github.com/google/tarpc/pull/380/commits/03f7236a489b86180e18a5f21800842af9179417)

tarpc/src/serde_transport.rs
```diff
@@ -437,6 +437,41 @@ pub mod unix {
...
+        fn as_ref(&self) -> &std::path::Path {
+            self.0.as_path()
+        }
+    }
+
+    impl Drop for TempSock {
+        fn drop(&mut self) {
+            let _ = std::fs::remove_file(&self.0);
+        }
+    }
```
tikue, line 474: Looks good! Can you also add some unit tests?

tarpc/src/serde_transport.rs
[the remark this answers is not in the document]
tikue: I think you can remove it for now; it's easy enough to add back if needed!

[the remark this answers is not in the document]
kbknapp: Removed in: (2143f75)[https://github.com/google/tarpc/pull/380/commits/2143f758f1a6c5043b43de28556fb05f0b09d77d]

tarpc/src/serde_transport.rs
kbknapp: I added a test to assert it's removed on Drop. If you'd me to add some tests to assert the naming parts as well I can do that, but it felt more heavy weight because of the randomness involved it would require either property testing or globs that would be more error prone.

tarpc/src/serde_transport.rs
```diff
@@ -433,22 +433,27 @@ pub mod unix {
     pub struct TempSock(std::path::PathBuf);
```
tikue, line 433: Nit, thinking about the naming: this is actually not creating the socket; it's choosing the place to create it, and then it's making sure it's cleaned up later. But it could be used for non-socket purposes, too (not that we intend it to be...).

What about something like FileCleanup or RemoveFileOnDrop?

tarpc/src/serde_transport.rs
```diff
@@ -433,22 +433,27 @@ pub mod unix {
...
+        }
+
+        /// A named socket that uses the name exactly as passed without allocating a new `String`
+        /// to add `.sock`, this results in `<tempdir>/<name>`
+        pub fn exact<S: AsRef<str>>(name: S) -> Self {
             let mut sock = std::env::temp_dir();
-            sock.push(format!("{}.sock", name.as_ref()));
+            sock.push(name.as_ref());
             Self(sock)
         }
```
tikue, line 447: this function should be easy enough to test? `assert!(temp_sock.ends_with(name));` or something?

tarpc/src/serde_transport.rs
kbknapp: Great point! Perhaps in this case sticking close to what it actually is would be a benefit? So something like `TempPathBuf` and just remove the `.sock` stuff. I'll do that real quick and add the tests for the naming as well.

kbknapp: Changed the name and removed `.sock` part of the names in: [fd0e243](https://github.com/google/tarpc/pull/380/commits/fd0e2434c0dfd7f1b276ef07ed64d77c8c942a13)

tarpc/src/serde_transport.rs
kbknapp: Added those tests along with some others including ensuring regular files are removed, or missing file errors are properly swallowed (such as if the `TempPathBuf` was never actually used to make a file/socket, Drop shouldn't panic).


[7da27a4](https://github.com/google/tarpc/pull/380/commits/7da27a45633c3d53f835d2a1aa2ac867dd297f8a)

tikue (approved):
Awesome, thanks so much!

dignifiedquire:
@tikue any chance of a release? I would love to start using this.

tikue:
Sorry for the delay! I'll get it released this week.

tikue:
It is now released, thank you for your patience! 🙏

dignifiedquire:
thank you!

PR #393, opened 2023-02-10
Add abitiy to store request context

### LINKED ISSUE #392: How to save transport context?
I'm trying to implement an MQTTv5 transport, and I have it more-or-less up and running, with fixed request and response topics.

There are two challanges I'm facing:
  1. Mqtt is not a one-to-one protocol, where each client will make a different connection to the server, but an aggregating one, where the messages of all clients will be seen through the same stream. The problem with this is that the message ids will collide. I'm thinking about sending randomized message ids instead of the sequential ones, that should mitigate the issue more-or-less. Is that correct?
  2. The way MQTTv5 request-response works is that the client publishing the message publishes two pieces of information together with the message: A correlation data (which will be the message id in this case), and a response topic where the client is awaiting the response. Which will be different in cases of requests coming from different clients, but the protocol allows for each message to contain different response topics as well. How can I save this information when decoding the ClientReq, so that when the Response arrives at the Sink I can publish the response to the correct topic? 
 
 Basically I'd need a way to save a transport context through the lifecycle of the request, so that it can be accessed when sending the response through the MQTT broker.

axos88:
Resolves #392 

If a transport is created that sinks/streams (C, Item/SinkItem), and that is used to create the BaseChannel, the transport will be able to generate a value (c:C) in the stream that will be opaquely forwarded to the Sink.

This can be used for example to differentiate in exaclty how a request arrived at the transport, for example if the transport's channel is aggregating multiple clients. It will be used mainly for MQTT, where the response topic is received together with the Request, and the Sink needs to be know where to publish the response.


axos88:
This is based on your branch containing the async trait features.

tarpc/src/server/in_flight_requests.rs
```diff
@@ -11,29 +11,40 @@ use tracing::Span;
...
-#[derive(Debug, Default)]
-pub struct InFlightRequests {
-    request_data: FnvHashMap<u64, RequestData>,
+#[derive(Debug)]
+pub struct InFlightRequests<C> {
+    request_data: FnvHashMap<u64, RequestData<C>>,
     deadlines: DelayQueue<u64>,
 }
 
+impl<C> Default for InFlightRequests<C> {
```
axos88, line 20: This is necessary, because the default Derive macro expects C: Default, but that is not strictly necessary.

axos88:
Signed the CLA.

axos88:
This currently breaks existing code. need to revise.

tikue:
Thanks for the PR! For now, I will continue the discussion on #392 and will wait to review this until I better understand the solution space.

tikue:
@axos88 since this has been open for a while, I'm going to close it for now. Let me know if you're still interested in discussing this—we could potentially revive our discussion on #392 (I would pretty much need to re-read everything to page back in the context).

axos88:
Hah yeah. I am currntly using my own fork,  would be great to get it merges, but currently have zero... or closer to the truth minus infinity resources to devote to this. Might reiterate on it in the future!

Issue #225, opened 2019-04-22
tls support


omac777:
When will TLS 1.3 support with tarpc come back?

tikue:
Hey, thanks for the question! Tarpc in its latest iteration is generic over the transport layer. My expectation is you should be able to wrap your transport in a TLS later without requiring any changes to the tarpc service API. Can you provide some details on what you've tried thus far?

omac777:
I haven't tried anything with tls so far.  I did try the example_service client and server and found them to be exactly what you claim tarpc to be....easy to use but constrained to exposing only one service from one server as there is no documentation demonstrating many services from one server.

In order for tarpc to become more popular, there would need to be more documentation and examples i.e. an example_service demonstrating tls 1.3 with it.
i.e. an example_service demonstrating exposing many services from one server.

As it stands, as user looking at the code, it is difficult for me to quickly add the necessary capability to become useful for me.  I am still finding golang's rpcx package easier to use to be quite honest.  I haven't found a rust equivalent to this easy to use rpc package.

tikue:
Thanks for the feedback! Adding an example showing TLS usage is a great idea.

Fuzen-py:
I tried wrapping tarpc_bincode around tokio_tls::TlsConnector, however tarpc_bincode requires tokio_tcp::TCPStream, maybe it should be around anything that implements `AsyncRead + AsyncWrite + Unpin` and not just TCPStream?

tikue:
Hmm, it is supposed to be generic over AsyncRead and AsyncWrite. Did you see [`Transport::from(t)`](https://github.com/google/tarpc/blob/5f6c3d7d989b7d7267c19d4b89c658450e86614e/bincode-transport/src/lib.rs#L117)?

Fuzen-py:
No I didn't, building on top of the hello example
```rust
    let server_addr = flags.value_of("server_addr").unwrap();
    let server_addr = server_addr
        .parse::<SocketAddr>()
        .unwrap_or_else(|e| panic!(r#"--serve_addr value "{}" invalid: {}"#, server_addr, e));

    let socket = tokio::net::TcpStream::connect(&server_addr).await?;
    let connector = native_tls::TlsConnector::builder().build().unwrap();
    let connector = tokio_tls::TlsConnector::from(connector);
    let socket = connector.connect("google.com", socket).await.unwrap();
    let transport = tarpc_bincode_transport::Transport::from(socket);

[...]

PR #398, opened 2023-03-12
Feature/tls over tcp example

cguentherTUChemnitz:
based on discussion of: https://github.com/google/tarpc/issues/225

i implemented a working example for tls_over_tpc, based on top of tarpc/examples/custom_transport and tokio-rustls examples and tests. Certificates are picked from tokio-rustls tests.


tikue (changes requested):
Thanks so much for contributing this example, this is awesome! I admit it mostly just looks like mumbo jumbo to me :) but the dependencies all look legit, and it's just an example, so I don't have any concerns.

I just have 2 tiny nits, take a look and then I think we can get this merged!

cguentherTUChemnitz:
I will update in the next week this PR. I have locally fiddled around also with client-auth and i was able to get it working correctly based on certs generating the same way as rustls is doing it for their tests:
https://github.com/rustls/rustls/tree/main/test-ca

tarpc/examples/tls_over_tcp.rs
```diff
@@ -0,0 +1,112 @@
...
+const PRIVATEKEY: &str = include_str!("certs/localhost-key.pem");
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    // -------------------- start here to setup tls tcp tokio stream --------------------------
+    // ref certs and loading from: https://github.com/tokio-rs/tls/blob/master/tokio-rustls/tests/test.rs
+    // ref basic tls server setup from: https://github.com/tokio-rs/tls/blob/master/tokio-rustls/examples/server/src/main.rs
+    let cert = certs(&mut BufReader::new(Cursor::new(CERT)))
+        .unwrap()
+        .drain(..)
```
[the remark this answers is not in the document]
cguentherTUChemnitz: I don't know. This was copied from rustls example. I migrated now to into_iter. Seems to work fine for me ;)

tarpc/examples/tls_over_tcp.rs
```diff
@@ -0,0 +1,112 @@
...
+        .map(rustls::Certificate)
+        .collect();
+    let mut keys = pkcs8_private_keys(&mut BufReader::new(Cursor::new(PRIVATEKEY))).unwrap();
+    let mut keys = keys.drain(..).map(rustls::PrivateKey);
```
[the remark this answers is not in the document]
cguentherTUChemnitz: I added a more generalized private key loading function encapsulating this.

tikue (approved):
Looks good, thanks so much!

tarpc/examples/certs/eddsa/end.chain
tikue, line 1: Where does this come from? It looks different from what's in https://github.com/tokio-rs/tls/blob/master/tokio-rustls/tests/end.chain — likewise for the other cert/chain/key files

tikue (changes requested):
Actually, I'd just like to understand a bit better how you generated the certificate files. Thanks so much!

tarpc/examples/certs/eddsa/end.chain
cguentherTUChemnitz: Good point. I have used the test-ca creation script from rustls, which generates a ca, as well as client and server certs. Since they are already also configured for client out i migrated to those ones. I linked the "how to create the certs" in the following comment line:
https://github.com/google/tarpc/pull/398/files#diff-ab7dfcbf41b879bf0b90f96552153640d5f8d9268cde12f3c20faa97cb97f6c7R33

tikue: Got it, thanks!

IzumiRaine:
Thank you so much, this is exactly what I needed right now :)

A minor comment on creating the client's root store: <https://github.com/google/tarpc/blob/bed85e282710a5d21444fd25a5c0f55796ffc889/tarpc/examples/tls_over_tcp.rs#L115-L122>
You can instead use `root_store.add` like when setting up the server's roots (it calls `RootCertStore::add_server_trust_anchors` internally), which is a bit simpler.

tikue:
Thanks @IzumiRaine ! A PR with the fix would be most welcome ☺️

IzumiRaine:
Oh, sure! I'll send a PR once I had a chance to play with it some more.

PR #399, opened 2023-03-16
Return transport errors to the caller

### LINKED ISSUE #389: No error details when exceeding of LengthDelimitedCodec's max frame length
We have an RPC service that takes a parameter of type `Vec<u8>`. Also, we use a [LengthDelimitedCodec](https://docs.rs/tokio-util/0.7.4/tokio_util/codec/length_delimited/struct.LengthDelimitedCodec.html) (as `framed_io`) in a serde transport (created with [tarpc::serde_transport::new](https://docs.rs/tarpc/latest/tarpc/serde_transport/fn.new.html)) in our client (created with [tarpc::client::new](https://docs.rs/tarpc/latest/tarpc/client/fn.new.html)).

If this service is called with the `Vec<u8>` parameter being larger than the LengthDelimitedCodec's max frame length (by default, this is 8MB), then this leads to the client disconnecting and returning an [RpcError](https://docs.rs/tarpc/latest/tarpc/client/enum.RpcError.html)::Disconnected (["the client disconnected from the server"](https://docs.rs/tarpc/latest/src/tarpc/client.rs.html#179)).

The error (message) is somewhat suboptimal because it doesn't give any indication on what the actual problem is (i.e., that the problem comes from exceeding the max frame length). In our case, it took several days of debugging (a production incident) to figure out what the problem is.

Would it maybe possible to improve the situation, e.g., by having a more detailed/specific error that directly points to the actual underlying problem?

bruwozniak:
resolves #389

This is just cherry picking relevant commits from https://github.com/tikue/tarpc/tree/gats and adding a test with a mock transport that errors out on sending.


tarpc/src/server.rs
```diff
@@ -343,11 +343,23 @@ pub enum ChannelError<E>
...
+    Write(#[source] E),
+    /// Could not flush the transport.
+    #[error("could not flush the transport")]
+    Flush(#[source] E),
+    /// Could not close the write end of the transport.
+    #[error("could not close the write end of the transport")]
+    Close(#[source] E),
+    /// Could not poll expired requests.
+    #[error("could not poll expired requests")]
     Timer(#[source] ::tokio::time::error::Error),
```
tikue: Is it possible to also add unit tests verifying the new error variants are returned under appropriate conditions?

tarpc/src/client.rs
```diff
@@ -790,6 +765,74 @@ mod tests {
...
+        let resx = resp.response().await;
+        assert_matches!(resx, Err(RpcError::Send(_)));
+        let error: anyhow::Error = resx.unwrap_err().into();
+        assert!(error
+            .root_cause()
+            .downcast_ref::<MockTransportError>()
+            .is_some());
+    }
+
+    struct MockTransport<S>(PhantomData<S>);
```
tikue, line 802: nit: how about a more descriptive name, like `AlwaysErrorTransport`?

tikue (changes requested):
Awesome, thanks so much! Just a couple of comments :)

tarpc/src/client.rs
bruwozniak: Totally!

tarpc/src/server.rs
bruwozniak: If you mean to check all the other enum cases using the `AlwaysErrorTransport`, then sure, will do.

tarpc/tests/compile_fail/must_use_request_dispatch.stderr
tikue, line 1: FYI: I just pushed these fixes because they were blocking other PRs

bruwozniak: Sure, it actually took me a good dozen minutes to actually understand what is broken and why 😄

bruwozniak:
@tikue Rebased and extended the test to all cases (I tried to compress it into one tabular test not to cause bloat). Please take anoter look.

tarpc/src/client.rs
```diff
@@ -764,72 +764,138 @@ mod tests {
...
-            to_dispatch,
-            cancellation,
-            next_request_id: Arc::new(AtomicUsize::new(0)),
-        };
-
-        let cx = &mut Context::from_waker(noop_waker_ref());
-        let (tx, mut rx) = oneshot::channel();
-        let resp = send_request(&mut channel, "hi", tx, &mut rx).await;
-        assert!(dispatch.as_mut().poll(cx).is_pending());
+        for (error, cause) in vec![
```
tikue, line 770: Ugh, I'm sorry about this, I just realized I had left my comment about test coverage on the wrong enum. This test, which is testing RpcError::Send, was totally fine as you had it before.

What I meant to say, but failed to, was that RpcError had some other variants added in this PR, and those new variants could use test coverage—are they returned in the proper circumstances? For example, Disconnected was replaced by Shutdown and Receive errors.

The additional test cases I had in mind:

- RpcError::Shutdown is returned when the dispatch task has shut down, preventing the client from:
    1. initiating an RPC
    2. completing an RPC
- An RPC should be completed with RpcError::Receive if the dispatch task receives an error when attempting to receive a message from the transport. Failing to receive is a permanent error, which shouldn't change, but the dispatch task should still send the more specific Receive error to all pending RPCs (of which there could be a few).

I've added comments to the specific places in the code that correspond to these test cases. I actually think RpcError::Receive is unused in this PR, so you can remove it if you'd like, and I can file an issue to add it back later.

tarpc/src/client.rs
```diff
@@ -158,28 +157,33 @@ impl<Req, Resp> Channel<Req, Resp> {
                 response_completion,
             })
             .await
-            .map_err(|mpsc::error::SendError(_)| RpcError::Disconnected)?;
+            .map_err(|mpsc::error::SendError(_)| RpcError::Shutdown)?;
```
tikue, line 160: Shutdown test case 1/2

tarpc/src/client.rs
```diff
@@ -188,24 +192,18 @@ pub enum RpcError {
...
         self.cancel = false;
         match response {
-            Ok(resp) => Ok(resp?.message?),
+            Ok(response) => response,
             Err(oneshot::error::RecvError { .. }) => {
                 // The oneshot is Canceled when the dispatch task ends. In that case,
                 // there's nothing listening on the other side, so there's no point in
                 // propagating cancellation.
-                Err(RpcError::Disconnected)
+                Err(RpcError::Shutdown)
```
tikue, line 206: Shutdown test case 2/2

tarpc/src/client.rs
```diff
@@ -158,28 +157,33 @@ impl<Req, Resp> Channel<Req, Resp> {
...
-    #[error("the client disconnected from the server")]
-    Disconnected,
+    #[error("the connection to the server was already shutdown")]
+    Shutdown,
+    /// The client failed to send the request.
+    #[error("the client failed to send the request")]
+    Send(#[source] Box<dyn std::error::Error + Send + Sync + 'static>),
+    /// An error occurred while waiting for the server response.
+    #[error("an error occurred while waiting for the server response")]
+    Receive(#[source] Box<dyn std::error::Error + Send + Sync + 'static>),
```
tikue, line 186: Receive seems unused, can either update pump_read to complete all pending RPCs with it, or just remove it (FWIW, I'll probably want to add it back later)

```rs
// client/in_flight_requests.rs
impl<Res> InFlightRequests<Res> {
    ...
    pub fn complete_all_requests<'a>(
        &'a mut self,
        mut result: impl FnMut() -> Res + 'a,
    ) -> impl Iterator<Item = Span> + 'a {
        self.deadlines.clear();
        self.request_data.drain().map(move |(_, request_data)| {
            let _ = request_data.response_completion.send(result());
            request_data.span
        })
    }
}
// client.rs
impl<Req, Resp, C> RequestDispatch<Req, Resp, C> {
    ...
    fn pump_read(
        mut self: Pin<&mut Self>,
        cx: &mut Context<'_>,
    ) -> Poll<Option<Result<(), ChannelError<C::Error>>>> {
        self.transport_pin_mut()
            .poll_next(cx)
            .map_err(|e| {
                let e = Arc::new(e);
                for span in self
                    .in_flight_requests()
                    .complete_all_requests(|| Err(RpcError::Receive(e.clone())))
                {
                    let _entered = span.enter();
                    tracing::info!("ReceiveError");
                }
                ChannelError::Read(e)
            })
            .map_ok(|response| {
                self.complete(response);
            })
    }
}
```

Note that this does require changing ChannelError::Read to contain an Arc<E>, which makes sense for the client, but does not seem to make sense for the server. This somewhat changes my opinion on unifying the two ChannelErrors. What do you think?

tikue (changes requested):

tarpc/src/client.rs
bruwozniak: OK, I added your suggestion, verbatim as I have not enough intuitions about the details of mechanics in this crate to spontaneously come up with this.
About unified ChannelError, I still think it's better to have one enum than two only very slightly different enums. One can either use the arced version for server side (performance penalty should be minimal) or have two variants i.e. `ServerReceive`/`ClientReceive`.

tarpc/src/client.rs
bruwozniak: No worries @tikue , it happens. I think now I added what you had in mind. I don't think it makes sense to delete what I already added though? It seems useful for the future and more tests usually don't hurt, WDYT?

tarpc/src/client/in_flight_requests.rs
```diff
@@ -93,6 +93,17 @@ impl<Res> InFlightRequests<Res> {
         false
     }
 
+    pub fn complete_all_requests<'a>(
```
tikue, line 96: Can you add a short doc comment?

```rs
/// Completes all requests using the provided function.
/// Returns Spans for all completes requests.
```

tarpc/src/client.rs
tikue: That sounds good to me, though I think these are really 5 different test cases (should be different `#[test]` fns). It's the match expression that leads me to think that — each error variant has different test logic.

I like the test coverage, but would prefer to put the test cases in different fns.

tikue (changes requested):
Just 2 minor things and then I think we're good to merge :)

tikue (approved):
Thanks so much for making this great improvement!

bruwozniak:
@tikue thanks you for very helpful guidance. May I ask for a release with this change so we can include it in our project?

tikue:
I published v0.32.0 immediately after merging this :)

PR #404, opened 2023-04-11
Simplify TLS example

IzumiRaine:
Just some minor simplifications as follow-up from my comment in PR #398; functionality should be unchanged.


tikue:
Amazing, thanks!

PR #407, opened 2023-09-11
Fix example in the README

cGuille:
After following the instructions from the README to replicate the example, I had a few warnings, as well as an error due to a missing tokio feature.

This PR contains the changes I made to fix those few issues.


tikue:
Thanks! Should any of these fixes also be applied to https://github.com/google/tarpc/blob/master/tarpc/examples/readme.rs?

cGuille:
I don't think so:
- There is no `Cargo.toml` in this directory, so the first issue about the missing tokio feature cannot happen.
- I could build without warning by using the `Cargo.toml`  of README.md and the sources from `tarpc/examples/readme.rs`.

So I think it's fine.

tikue:
Hm, looks like I've got some bitrot to take care of! Will get back to this soon.

tikue:
Thanks again! Sorry for the delay!

cGuille:
No worry, thank you!

PR #410, opened 2023-11-15
Update OpenTelemetry Packages in Example App

akevinge:
This just updates the OpenTelemetry setup in the example app. For beginners like me, a mention in the docs that `RUST_LOG`  is required to be set to see the service in Jaeger would be nice (not sure if that belongs in this repo though).

### DIFF
--- a/example-service/Cargo.toml
+++ b/example-service/Cargo.toml
@@ -18,14 +18,15 @@ anyhow = "1.0"
 clap = { version = "3.0.0-rc.9", features = ["derive"] }
 log = "0.4"
 futures = "0.3"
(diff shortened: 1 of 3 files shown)


tikue:
Thanks! Feel free to update the readme's section on Jaeger to clarify RUST_LOG details.

tikue:
Hm, looks like there's a test failure?

akevinge:
Interesting. I was unable to produce the error locally even after updating my toolchain to latest stable. The test also seems unrelated to the modified code. I wonder if the test is flaky? Would it be possible to rerun the workflow?

PR #411, opened 2023-12-06
Allow to create tarpc server on existing listener

SabatierBoris:
When you want to use systemd.socket or systemfd for allocate the socket (TCP or UNIX) you need to be able to create a server on already existing listener

### DIFF
--- a/tarpc/src/serde_transport.rs
+++ b/tarpc/src/serde_transport.rs
@@ -364,7 +364,19 @@ pub mod unix {
         Codec: Serializer<SinkItem> + Deserializer<Item>,
         CodecFn: Fn() -> Codec,
     {
-        let listener = UnixListener::bind(path)?;
+        listen_on(UnixListener::bind(path)?, codec_fn).await
+    }
+
+    /// Wrap accepted connections from `listener` in Unix Domain Socket transports.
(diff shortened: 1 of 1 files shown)


tikue:
Thanks, and apologies for the review delay! Can you add unit tests for the new functions? Other than that, looks great.

SabatierBoris:
> Thanks, and apologies for the review delay! Can you add unit tests for the new functions? Other than that, looks great.

Hi, no problem.

I add some tests based on existing TCP and UDS tests.

tarpc/src/serde_transport.rs
```diff
@@ -669,4 +713,25 @@ mod tests {
         assert_matches!(transport.next().await, None);
         Ok(())
     }
+
+    #[cfg(all(unix, feature = "unix"))]
+    #[tokio::test]
+    async fn uds_on_existing_transport() -> io::Result<()> {
+        use super::unix;
+        use super::*;
```
tikue, line 721: Just curious, what else was needed besides `super::unix`?

tikue: Ah, I think the imports on line 565 just need to be updated to include StreamExt and SinkExt. I see another test already imports super::*, so I'll just update this in a separate commit.

tikue (approved):

PR #414, opened 2024-01-09
Fix cfg flag for tcp tests

SabatierBoris:
# Issue

tcp test for serde_transport isn't run due to wrong `cfg` flag

# Fix

Use same `cfg` flag for test as for the tcp module  : https://github.com/google/tarpc/blob/master/tarpc/src/serde_transport.rs#L120

### DIFF
--- a/tarpc/src/serde_transport.rs
+++ b/tarpc/src/serde_transport.rs
@@ -655,7 +655,7 @@ mod tests {
         );
     }
 
-    #[cfg(tcp)]
+    #[cfg(feature = "tcp")]
     #[tokio::test]
     async fn tcp() -> io::Result<()> {
         use super::tcp;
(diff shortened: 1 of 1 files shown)


tikue:
Thanks!

tarpc/src/serde_transport.rs
```diff
@@ -674,12 +674,12 @@ mod tests {
...
     }
 
-    #[cfg(tcp)]
+    #[cfg(feature = "tcp")]
     #[tokio::test]
     async fn tcp_on_existing_transport() -> io::Result<()> {
         use super::tcp;
 
-        let transport = TcpListener::bind("0.0.0.0:0").await?;
+        let transport = tokio::net::TcpListener::bind("0.0.0.0:0").await?;
```
tikue, line 682: Was this change needed?

Issue #415, opened 2024-01-10
On client's read errors quick subsequent requests may wait for response indefinitely


altkdf:
If I understand correctly, the dispatch task is terminated on a receiving error in transport (from `pump_read()`) and subsequent requests return in 
```rust
        self.to_dispatch
            .send(DispatchRequest {
                ctx,
                span,
                request_id,
                request,
                response_completion,
            })
            .await
            .map_err(|mpsc::error::SendError(_)| RpcError::Shutdown)?;
```

But for me it does sometime happen that if the request is filed immediately after the error, then 
```rust
        self.to_dispatch
            .send
```
does not error and waits indefinitely in `response_guard.response().await`. Not sure why this is happening, since I would expect that all `DispatchRequest`s would be dropped and also the waiters would notice that and return an error. But it seems that there is a race condition?

Unfortunately, I couldn't produce a minimal working example. A small example "just works".

If I add `self.pending_requests_mut().close();` to `pump_read()`, I can't reproduce the issue anymore. Not sure if that is a solution or it just masks the issue.

```rust
    fn pump_read(
        mut self: Pin<&mut Self>,
        cx: &mut Context<'_>,
    ) -> Poll<Option<Result<(), ChannelError<C::Error>>>> {
        self.transport_pin_mut()
            .poll_next(cx)
            .map_err(|e| {
                let e = Arc::new(e);
                for span in self
                    .in_flight_requests()
                    .complete_all_requests(|| Err(RpcError::Receive(e.clone())))
                {
                    let _entered = span.enter();
                    tracing::info!("ReceiveError");
                }
                self.pending_requests_mut().close();
                ChannelError::Read(e)
            })
            .map_ok(|response| {
                self.complete(response);
            })
    }
```

tikue:
Thanks for raising this issue! I'll try to take a closer look this week, but also, feel free to send a PR for review.

tikue:
Could I see your code where the problem occurs? I'm curious how the RequestDispatch is being polled. When it hits an error, is it dropped?

tikue:
BTW, I was able to refactor an existing test to consistently trigger this race: https://github.com/google/tarpc/compare/master...request_dispatch_race

tikue:
Can you see if this branch fixes the problem for you? https://github.com/tikue/tarpc/tree/request_dispatch_race

altkdf:
Oh, true, you're right - the code should work just fine like that!

[...]

PR #416, opened 2024-01-11
close request receiving channel on errors immediately

altkdf:
Fixes the problem described in https://github.com/google/tarpc/issues/415.

### DIFF
--- a/tarpc/src/client.rs
+++ b/tarpc/src/client.rs
@@ -345,13 +345,25 @@ where
             .poll_next(cx)
             .map_err(|e| {
                 self.pending_requests_mut().close();
+                let e = Arc::new(e);
                 // drain sender messages after closing the channel
                 loop {
-                    if matches!(self.pending_requests_mut().poll_recv(cx), Poll::Ready(None)) {
-                        break;
+                    match self.pending_requests_mut().poll_recv(cx) {
+                        // channel is open or a spurious failure
+                        Poll::Pending => continue,
+                        // a request is available
+                        Poll::Ready(Some(request)) => {
+                            // propagate the error to the waiter if it has not
+                            // hung up yet
+                            let _ = request
+                                .response_completion
+                                .send(Err(RpcError::Receive(e.clone())));
+                        }
+                        // no requests, no pending `Permits`, and the channel is
+                        // closed
+                        Poll::Ready(None) => break,
                     }
                 }
-                let e = Arc::new(e);
                 for span in self
                     .in_flight_requests()
                     .complete_all_requests(|| Err(RpcError::Receive(e.clone())))
--- a/tarpc/src/client.rs
+++ b/tarpc/src/client.rs
@@ -353,6 +353,12 @@ where
                     tracing::info!("ReceiveError");
                 }
                 self.pending_requests_mut().close();
+                // drain sender messages after closing the channel
+                loop {
+                    if matches!(self.pending_requests_mut().poll_recv(cx), Poll::Ready(None)) {
+                        break;
+                    }
+                }
                 ChannelError::Read(e)
             })
             .map_ok(|response| {
--- a/tarpc/src/client.rs
+++ b/tarpc/src/client.rs
@@ -352,6 +352,7 @@ where
                     let _entered = span.enter();
                     tracing::info!("ReceiveError");
                 }
+                self.pending_requests_mut().close();
                 ChannelError::Read(e)
             })
             .map_ok(|response| {
--- a/tarpc/src/client.rs
+++ b/tarpc/src/client.rs
@@ -344,6 +344,13 @@ where
         self.transport_pin_mut()
             .poll_next(cx)
             .map_err(|e| {
+                self.pending_requests_mut().close();
+                // drain sender messages after closing the channel
+                loop {
+                    if matches!(self.pending_requests_mut().poll_recv(cx), Poll::Ready(None)) {
+                        break;
+                    }
+                }
                 let e = Arc::new(e);
                 for span in self
                     .in_flight_requests()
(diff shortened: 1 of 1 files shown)


altkdf:
I signed the CLA but don't know how to rerun the CLA job.

altkdf:
Clicking "rescan" gives me a "400. That’s an error. That’s all we know."

tarpc/src/client.rs
```diff
@@ -352,6 +352,7 @@ where
                     let _entered = span.enter();
                     tracing::info!("ReceiveError");
                 }
+                self.pending_requests_mut().close();
```
tikue, line 355: The thing that's confusing me is, pending requests should close once the request dispatch is dropped. Can you should me your code that polls the request dispatch? What does the code do after `RequestDispatch::poll` returns an error?

altkdf: According to the [documentation](https://docs.rs/tokio/latest/tokio/sync/mpsc/index.html#clean-shutdown) of `tokio` it is possible to have messages in the channel after dropping the `Receiver`. This does not happen on each run but maybe once in 10 or 100 times.  My test currently produces the following log.
```
2024-01-23T22:14:29.394867Z  INFO tarpc::client: dropping request dispatch
2024-01-23T22:14:29.394871Z  INFO RPC{rpc.deadline=2024-01-23T22:14:39.394769825Z otel.kind="client" otel.name="TarpcCspVault.idkg_gen_dealing_encryption_key_pair" rpc.trace_id=00}: tarpc::client: sending request to dispatch
2024-01-23T22:14:29.394973Z TRACE tarpc::server: Expired requests: Closed, Inbound: Closed
2024-01-23T22:14:29.395005Z TRACE tarpc::server: poll_flush
2024-01-23T22:14:29.395019Z TRACE tokio_util::codec::framed_impl: flushing framed transport
2024-01-23T22:14:29.395032Z TRACE tokio_util::codec::framed_impl: framed transport flushed
2024-01-23T22:14:29.395025Z  WARN tarpc::client: Connection broken: could not read from the transport

Caused by:
    frame size too big
2024-01-23T22:14:29.395086Z  INFO RPC{rpc.deadline=2024-01-23T22:14:39.394769825Z otel.kind="client" otel.name="TarpcCspVault.idkg_gen_dealing_encryption_key_pair" rpc.trace_id=00}: tarpc::client: sending request to dispatch done; waiting for response
test rpc_connection::should_unfortunately_be_dead_after_response_from_server_cannot_be_received_by_client_because_too_large has been running for over 60 seconds
```

Our code makes use of the `tarpc` `Client` as a black box by calling its generated methods using `tokio`'s `block_on`. Here's an [example](https://github.com/dfinity/ic/blob/master/rs/crypto/internal/crypto_service_provider/src/vault/remote_csp_vault/tarpc_csp_vault_client.rs#L271).

tikue: Thanks! Will review your code and get back to you.

(It shouldn't be a problem if there are messages in the channel when the channel is dropped — the messages in the channel contain the sender side of oneshot channels used to send individual RPC responses. When the channel is dropped, the senders are also dropped, and that should cause the receiver side to receive an error.)

altkdf: Thank you!

I'm not sure if the channel is actually dropped. The only thing that is dropped (IIUC) is the receiver part of the dispatch channel and some messages from the channel. But the channel itself, again IIUC, is just a synchronized queue that is shared by both sender and receiver. The sender part is not dropped until we drop the client, right, it only changes to an errored state to prevent sending further messages to the channel? My hypothesis so far was that:
1) A dispatch request (which contains the sender part of the response channel) is sent to the request dispatcher.
2) The caller waits on the receiver end of the response channel.
3) The dispatcher receiver is being dropped but the channel is not properly cleared, so the dispatch request along with the response sender handle is not dropped.
4) Client waits indefinitely.

WDYT @tikue?

tikue: Ah, that's really interesting and makes a lot of sense, thanks!

tikue, line 355: I think there could be pending requests after closing the receiver. Maybe those requests should be completed as well, similar to what's done with in_flight_requests (immediately above)? Something like this: https://github.com/google/tarpc/blob/876fd4724b4ff051d8631be8bbc892023bd98e71/tarpc/src/client/in_flight_requests.rs#L102C19-L102C69

altkdf: I pushed some commits, PTAL.

tarpc/src/client.rs
```diff
@@ -344,7 +344,26 @@ where
         self.transport_pin_mut()
             .poll_next(cx)
             .map_err(|e| {
+                self.pending_requests_mut().close();
                 let e = Arc::new(e);
+                // drain sender messages after closing the channel
+                loop {
+                    match self.pending_requests_mut().poll_recv(cx) {
+                        // channel is open or a spurious failure
+                        Poll::Pending => continue,
```
tikue, line 353: Oh, acually, the channel returns Pending when it's empty, I think. https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.poll_recv

So this loop can break on `Poll::Pending | Poll::Ready(None)`.

tikue: Oh, I see now why you did it this way — basically, the receiver is never supposed to return `Poll::Pending` after closure. Interesting — I guess this is good then!

tikue: Actually, I think there's some subtlety here: reading the docs for [close](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.close), it looks like outstanding Permits are still allowed to send into the channel. But there's no guarantee when the send will happen, so this loop is actually a blocking call. Since this code path is nonblocking, I think that's a problem we have to solve.

What about spawning a task that drains the channel?

On a separate note, I think there are other errors that could occur besides the error in `pump_read`, and I hadn't made any attempt to drain the requests for those paths. And likewise, errors in other areas (like in pump_write) will not result in the receiver being drained. What if all of this draining was moved into a drop impl for RequestDispatch?

tikue:
Obsoleted by https://github.com/google/tarpc/pull/423. Thanks so much for identifying this problem as well as the solution!

PR #417, opened 2024-01-13
cleanup std usages / procedural macro hygiene

OMGeeky:
Directly using things like Result and Option without namespace inside the generated code without using the full namespace can lead to problems when the caller has renamed those usages and provided his own with different signatures (often done with the Result type).

See the [Rust Reference](https://doc.rust-lang.org/reference/procedural-macros.html#procedural-macro-hygiene) on this.

One example that would create compiler errors without this fix:
```Rust
use std::result::Result as OldResult;
type Result<T> = OldResult<T,u16>;
use std::option::Option as OldOption;
type Option = OldOption<u16>;

#[tarpc::service]
pub trait World {
    async fn hello(name: String) -> String;
}
```

### DIFF
--- a/plugins/src/lib.rs
+++ b/plugins/src/lib.rs
@@ -32,8 +32,8 @@ use syn::{
 macro_rules! extend_errors {
     ($errors: ident, $e: expr) => {
         match $errors {
-            ::std::result::Result::Ok(_) => $errors = ::std::result::Result::Err($e),
-            ::std::result::Result::Err(ref mut errors) => errors.extend($e),
+            Ok(_) => $errors = Err($e),
+            Err(ref mut errors) => errors.extend($e),
         }
     };
 }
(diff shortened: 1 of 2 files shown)


tikue:
Thank you!

plugins/src/lib.rs
```diff
@@ -431,7 +431,7 @@ impl<'a> ServiceGenerator<'a> {
                 }
 
                 async fn serve(self, ctx: tarpc::context::Context, req: #request_ident)
-                    -> Result<#response_ident, tarpc::ServerError> {
+                    -> ::std::result::Result<#response_ident, tarpc::ServerError> {
                     match req {
                         #(
                             #request_ident::#camel_case_idents{ #( #arg_pats ),* } => {
```
tikue, line 437: what about the use of `Ok` and `Some` etc.? Do you want to clean those up, too?

OMGeeky: Sure, I just missed some of those, but it makes sense.

OMGeeky:
I Added some more and I think I found everything that I missed in the previous one.

OMGeeky:
I did find some more, like the unreachable macro. I found an easy way to check for this and will create a test that checks for any external usages.

OMGeeky:
One thing I also changed was all the usages of `tarpc` to `::tarpc` because if you rename your import you can still access it with those two `::` in front of the crate name, by the name it was under originally.

tikue:
Thanks so much for fixing this!

PR #422, opened 2024-01-30
Use Instant (monotonic) time instead of SystemTime

mislavn:
This continues on the #367 pull request.

Tarpc should use relative time, especially since RPC serialization no longer uses SystemTime but Duration.

With this commit Tarpc uses Instant time and on the edges converts Instant time to SystemTime for the Opentelemetry span's.

### DIFF
--- a/tarpc/src/client.rs
+++ b/tarpc/src/client.rs
@@ -11,7 +11,9 @@ pub mod stub;
 
 use crate::{
     cancellations::{cancellations, CanceledRequests, RequestCancellation},
-    context, trace, ChannelError, ClientMessage, Request, Response, ServerError, Transport,
+    context, trace,
+    util::TimeUntil,
+    ChannelError, ClientMessage, Request, Response, ServerError, Transport,
 };
 use futures::{prelude::*, ready, stream::Fuse, task::*};
 use in_flight_requests::InFlightRequests;
@@ -24,6 +26,7 @@ use std::{
         atomic::{AtomicUsize, Ordering},
         Arc,
     },
+    time::SystemTime,
 };
 use tokio::sync::{mpsc, oneshot};
 use tracing::Span;
@@ -117,7 +120,7 @@ impl<Req, Resp> Channel<Req, Resp> {
         skip(self, ctx, request_name, request),
         fields(
             rpc.trace_id = tracing::field::Empty,
-            rpc.deadline = %humantime::format_rfc3339(ctx.deadline),
+            rpc.deadline = %humantime::format_rfc3339(SystemTime::now() + ctx.deadline.time_until()),
             otel.kind = "client",
             otel.name = request_name)
         )]
--- a/tarpc/src/server.rs
+++ b/tarpc/src/server.rs
@@ -9,7 +9,9 @@
 use crate::{
     cancellations::{cancellations, CanceledRequests, RequestCancellation},
     context::{self, SpanExt},
-    trace, ChannelError, ClientMessage, Request, Response, ServerError, Transport,
+    trace,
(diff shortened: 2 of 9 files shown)


tikue:
I like this, thanks! I'll review this week.

mislavn:
Thanks for the quick reply!

I just want to add a bit of context. We use Tarpc to communicate with an embedded device that doesn't have an internal clock. During runtime we update the clock and that may cause RPC timeouts. This PR eliminates the issue.

I haven't added any new tests because I couldn't figure out an easy way of mocking the SystemTime for testing.

tikue:
Sounds like a cool application! Out of curiosity, what are you building?

tarpc/src/util.rs
```diff
@@ -7,22 +7,22 @@
...
-    time::{Duration, SystemTime},
+    time::{Duration, Instant},
 };
 
 #[cfg(feature = "serde1")]
 #[cfg_attr(docsrs, doc(cfg(feature = "serde1")))]
 pub mod serde;
 
-/// Extension trait for [SystemTimes](SystemTime) in the future, i.e. deadlines.
+/// Extension trait for [Instant's](Instant) in the future, i.e. deadlines.
```
tikue, line 17: Nit: no apostrophe

```suggestion
/// Extension trait for [Instants](Instant) in the future, i.e. deadlines.
```

tarpc/src/client.rs
```diff
@@ -117,7 +118,7 @@ impl<Req, Resp> Channel<Req, Resp> {
         skip(self, ctx, request_name, request),
         fields(
             rpc.trace_id = tracing::field::Empty,
-            rpc.deadline = %humantime::format_rfc3339(ctx.deadline),
+            rpc.deadline = %humantime::format_rfc3339(SystemTime::now() + (ctx.deadline - Instant::now())),
```
tikue, line 121: nit: I think you can use `tarpc::util::TimeUntil` here
```suggestion
            rpc.deadline = %humantime::format_rfc3339(SystemTime::now() + ctx.deadline.time_until()),
```

tarpc/src/server.rs
```diff
@@ -348,7 +356,7 @@ where
         let span = info_span!(
             "RPC",
             rpc.trace_id = %request.context.trace_id(),
-            rpc.deadline = %humantime::format_rfc3339(request.context.deadline),
+            rpc.deadline = %humantime::format_rfc3339(SystemTime::now() + (request.context.deadline - Instant::now())),
```
tikue, line 359:
```suggestion
            rpc.deadline = %humantime::format_rfc3339(SystemTime::now() + request.context.deadline.time_until()),
```

tikue (approved):

tarpc/src/server.rs
mislavn: Done, thanks for the feedback.

tarpc/src/util.rs
mislavn: Done.

tarpc/src/client.rs
mislavn: Done.

mislavn:
> Sounds like a cool application! Out of curiosity, what are you building?

We are building X-ray detectors. Tarpc with channels is used for communication between our "microservices" on the detector, while for the public API facing the PC we use Cap'n proto but we still use Tarpc for testing and internal tools.

tikue:
There are some failures — I think there are just some missing imports for the TimeUntil trait.
image

mislavn:
It should be OK now, I verified it locally.

tikue:
Thanks again, I think this is a great change!

PR #427, opened 2024-02-03
update syn 2.0

getong:


plugins/src/lib.rs
```diff
@@ -172,27 +172,29 @@ impl Parse for DeriveSerde {
                 );
                 continue;
             }
-            match meta.lit {
-                Lit::Bool(LitBool { value: true, .. }) if cfg!(feature = "serde1") => {
-                    result = result.and(Ok(Some(true)))
-                }
-                Lit::Bool(LitBool { value: true, .. }) => {
-                    extend_errors!(
+            if let Expr::Lit(expr_lit) = &meta.value {
```
tikue, line 175: what if meta.value does not match Expr::Lit? I think there should be an error in that case.

On HEAD, the following service won't compile, but it does with this change:

```rs
#[tarpc::service(derive_serde = loop {})]
pub trait World {
    /// Returns a greeting for name.
    async fn hello(name: String) -> String;
}
```

Error on HEAD:

```
error: expected literal
  --> example-service/src/lib.rs:12:33
   |
12 | #[tarpc::service(derive_serde = loop {})]
```

Also, I think this means we're missing a test case. Would you like to add one to tarpc/tests/compile_fail? I'm also happy to add it too.

tikue:
Thanks very much! Left one review comment regarding literal validation in `derive_serde`

plugins/src/lib.rs
```diff
@@ -172,27 +172,29 @@ impl Parse for DeriveSerde {
...
+                    Lit::Bool(LitBool { value: true, .. }) => {
+                        extend_errors!(
+                            result,
+                            syn::Error::new(
+                                meta.span(),
+                                "To enable serde, first enable the `serde1` feature of tarpc"
+                            )
+                        );
+                    }
+                    Lit::Bool(LitBool { value: false, .. }) => result = result.and(Ok(Some(false))),
```
tikue, line 189: Here's a refactoring that catches the non-literal error:

```suggestion
            let Expr::Lit(expr_lit) = &meta.value else {
                extend_errors!(
                    result,
                    syn::Error::new(
                        meta.value.span(),
                        "`derive_serde` expects a literal of type `bool`"
                    )
                );
                continue;
            };
            match expr_lit.lit {
                Lit::Bool(LitBool { value: true, .. }) if cfg!(feature = "serde1") => {
                    result = result.and(Ok(Some(true)))
                }
                Lit::Bool(LitBool { value: true, .. }) => {
                    extend_errors!(
                        result,
                        syn::Error::new(
                            meta.span(),
                            "To enable serde, first enable the `serde1` feature of tarpc"
                        )
                    );
                }
                Lit::Bool(LitBool { value: false, .. }) => result = result.and(Ok(Some(false))),
```

plugins/src/lib.rs
tikue: Disregard that last question — I added a test case for this in https://github.com/google/tarpc/commit/cef01a1d590923a4577292d49c7a880ac5e17eb1 (feel free to TRYBUILD=overwrite if you update the compile error to be more specific)

getong:
clap, tokio-serde, console-subscriber, all of them upgrade to new version, since they use syn 1.0, which cause the ci error.
After upgrading new version, the ci is ok now.

tarpc/Cargo.toml
```diff
@@ -53,7 +53,7 @@ tarpc-plugins = { path = "../plugins", version = "0.13" }
 thiserror = "1.0"
 tokio = { version = "1", features = ["time"] }
 tokio-util = { version = "0.7.3", features = ["time"] }
-tokio-serde = { optional = true, version = "0.8" }
+tokio-serde = { optional = true, git = "https://github.com/tamird/tokio-serde", branch = "remove-educe", features = ["json", "bincode"] }
```
tikue, line 56: I'd rather depend on tokio-serde even if it pulls in 2 versions of syv. Consider a PR to tokio-serde?

tikue (changes requested):
Looks good other than the tokio-serde change

getong:
https://github.com/carllerche/tokio-serde/pull/31

I found this pr， but it does not be merged.

getong:
@tikue 
Everything is ok now

tikue (approved):

tikue:
Thank you!

PR #435, opened 2024-02-08
 Add support for #[cfg(…)] attributes on rpc trait definition

paulfariello:
Currently it's not possible to use a `#[cfg(…)]` attribute on a method.
```rust
#[tarpc::service]
pub trait World {
    #[cfg(feature = "hello")]
    async fn hello() -> String;
}
```
Would result in:
```
error[E0599]: no function or associated item named `hello` found for trait `World`
 --> rpc/src/lib.rs:8:14
  |
5 |   pub trait World {
  |  ___________-
6 | |     /// Returns a greeting for name.
7 | |     #[cfg(feature = "hello")]
8 | |     async fn hello() -> String;
  | |             -^^^^^ function or associated item not found in `World`
  | |_____________|
  |

For more information about this error, try `rustc --explain E0599`.
error: could not compile `rpc` (lib) due to previous error
```

This PR propose to simply propagate *all* attributes to every enum, and impl that are linked to a given rpc.
Obviously this might cause issue with attributes that are only supposed to be used on function.

Would this kind of solution be acceptable? If not, what kind of solution would you consider to enable `#[cfg(…)]` usage?

### DIFF
--- a/plugins/src/lib.rs
+++ b/plugins/src/lib.rs
@@ -22,8 +22,8 @@ use syn::{
     parse_macro_input, parse_quote,
     spanned::Spanned,
     token::Comma,
-    Attribute, Expr, FnArg, Ident, Lit, LitBool, MetaNameValue, Pat, PatType, ReturnType, Token,
-    Type, Visibility,
+    AttrStyle, Attribute, Expr, FnArg, Ident, Lit, LitBool, MetaNameValue, Pat, PatType,
+    ReturnType, Token, Type, Visibility,
 };
 
 /// Accumulates multiple errors into a result.
@@ -289,6 +289,23 @@ pub fn service(attr: TokenStream, input: TokenStream) -> TokenStream {
         vis,
         args,
         method_attrs: &rpcs.iter().map(|rpc| &*rpc.attrs).collect::<Vec<_>>(),
+        method_cfgs: &rpcs
+            .iter()
+            .map(|rpc| {
+                rpc.attrs
+                    .iter()
+                    .filter(|att| {
+                        att.style == AttrStyle::Outer
+                            && match &att.meta {
+                                syn::Meta::List(syn::MetaList { path, .. }) => {
+                                    path.get_ident() == Some(&Ident::new("cfg", rpc.ident.span()))
+                                }
+                                _ => false,
+                            }
+                    })
+                    .collect::<Vec<_>>()
+            })
+            .collect::<Vec<_>>(),
         method_idents: &methods,
         request_names: &request_names,
         attrs,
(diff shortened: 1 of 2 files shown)


tikue:
Hi, thanks for working to fix this! Interesting, I hadn't thought about supporting `#[cfg]` on RPC methods. I'm curious (though doesn't affect what I think about this PR), what's your use case?

What about a more selective fix, to basically *only* add attributes to the enum variants if if they are `[cfg()]` attributes? Right now you can see there is a failing clippy check due to doc comments (`/// like this`) being added to match arms.

paulfariello:
Our main use case is to disable a specific RPC when building our app for production. This RPC is mainly used for simulation of a specific behavior.

Sure, trying to look for specific `#[cfg(…)]` attributes seems quite reasonable, I'll work on it.

plugins/src/lib.rs
```diff
@@ -289,6 +289,23 @@ pub fn service(attr: TokenStream, input: TokenStream) -> TokenStream {
...
+                    .filter(|att| {
+                        att.style == AttrStyle::Outer
+                            && match &att.meta {
+                                syn::Meta::List(syn::MetaList { path, .. }) => {
+                                    path.get_ident() == Some(&Ident::new("cfg", rpc.ident.span()))
+                                }
+                                _ => false,
+                            }
+                    })
+                    .collect::<Vec<_>>()
```
tikue, line 306: nit: since this might not be super obvious to a reader, perhaps move it into a function called something like `CollectCfgAttrs`?

tikue:
Could you also add a test case in https://github.com/google/tarpc/blob/master/plugins/tests/service.rs?

plugins/src/lib.rs
paulfariello: Done with 8d9e433e908bdde54d5e781f5af69aad02d28e07

plugins/tests/service.rs
```diff
@@ -56,6 +56,27 @@ fn raw_idents() {
     }
 }
 
+#[test]
+fn service_with_cfg_rpc() {
+    #[tarpc::service]
+    trait Foo {
+        async fn foo(s: String, i: i32) -> (String, i32);
+        #[cfg(feature = "inexistant")]
```
tikue, line 64: how about using `#[cfg(not(test))]` to make sure it actually gets compiled out? And then leave out the fn below?

```suggestion
        #[cfg(not(test))]
```

paulfariello: Since `#[cfg(…)]` were not propagated the test was failing anyway, but you are right, leaving out the function in the impl is better.
Using `#[cfg(not(test))]` is a great idea!

I just fixed the commits according to your comment.

tikue:
Thanks very much for doing this!

Issue #436, opened 2024-02-22
Add a way to derive Clone for the generated Request and Response types


ShaneMurphy2:
Currently `tarpc::service` creates request and response types that are not clone. This makes it difficult it difficult to test and mock. It would be nice if the service macro used an attribute macro to control adding derives for things like `PartialEq`, `Clone`, or `Hash`.
Is there any reason not to do this? If there isn't, I can take a stab at it.

tikue:
Thanks for the suggestion! I agree this could be supported.

It'd be nice to replace the `derive_serde` arg with something generic enough to support this new use case. Maybe something like this would be possible:

```rust
#[tarpc::service(derive(tarpc::serde::Serialize, tarpc::serde::Deserialize, Clone, Hash))]
service World {
    async fn hello(message: String) -> String;
}
```

ShaneMurphy2:
@tikue Got a first cut up on a branch: https://github.com/ShaneMurphy2/tarpc/tree/customizable-derives, let me know what you think so far.

tikue:
Looks nice! It'd be nice for `derive_serde` and `derive` to be mutually exclusive, so that people are encouraged to make the switch. In particular:

- `derive_serde` and `derive` at the same time should be an error
- `derive` should only derive Serialize/Deserialize when they are explicitly listed
- `derive_serde` is deprecated - output a warning when it is used, and suggest instead using either `derive = []` to disable, or `derive = [Serialize, Deserialize]` to explicitly enable
- for backwards compatibility, absence of `derive` can derive Serialize/Deserialize by default.

ShaneMurphy2:
@tikue PR is up!

PR #438, opened 2024-02-23
Support custom derives on Request and Response

ShaneMurphy2:
Replace the bespoke "derive_serde" with a more flexible "derive = [<path>, <path>, ...]" form.

Deprecate the old "derive_serde" form, and emit deprecation warnings.

Add trybuild tests in a separate test crate to ensure the deprecation, features, and error messages all behave as expected. This technique of a separate trybuild test crate is borrowed from tokio/tests-build. Update CI to run this new crate's tests under the serde matrix.

### DIFF
--- a/plugins/src/lib.rs
+++ b/plugins/src/lib.rs
@@ -22,7 +22,7 @@ use syn::{
     parse_macro_input, parse_quote,
     spanned::Spanned,
     token::Comma,
-    AttrStyle, Attribute, Expr, FnArg, Ident, Lit, LitBool, MetaNameValue, Pat, PatType,
+    AttrStyle, Attribute, Expr, FnArg, Ident, Lit, LitBool, MetaNameValue, Pat, PatType, Path,
     ReturnType, Token, Type, Visibility,
 };
 
@@ -141,14 +141,37 @@ impl Parse for RpcMethod {
     }
 }
 
-// If `derive_serde` meta item is not present, defaults to cfg!(feature = "serde1").
-// `derive_serde` can only be true when serde1 is enabled.
-struct DeriveSerde(bool);
+#[derive(Default)]
+struct DeriveMeta {
+    derive: Option<Derive>,
+    warnings: Vec<TokenStream2>,
+}
 
-impl Parse for DeriveSerde {
+impl DeriveMeta {
+    fn with_derives(mut self, new: Vec<Path>) -> Self {
+        match self.derive.as_mut() {
+            Some(Derive::Explicit(old)) => old.extend(new),
+            _ => self.derive = Some(Derive::Explicit(new)),
+        }
+
+        self
+    }
+}
+
+enum Derive {
+    Explicit(Vec<Path>),
+    Serde(bool),
+}
+
+impl Parse for DeriveMeta {
     fn parse(input: ParseStream) -> syn::Result<Self> {
-        let mut result = Ok(None);
+        let mut result = Ok(DeriveMeta::default());
+
+        let mut derives = Vec::new();
         let mut derive_serde = Vec::new();
+        let mut has_derive_serde = false;
+        let mut has_explicit_derives = false;
+
         let meta_items = input.parse_terminated(MetaNameValue::parse, Comma)?;
         for meta in meta_items {
             if meta.path.segments.len() != 1 {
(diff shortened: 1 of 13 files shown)


ShaneMurphy2:
Not sure if I did the CI modification correctly, please take a look.

ShaneMurphy2:
Addresses #436

ShaneMurphy2:
Looks like one thing I still need to iron out (besides the CLA 😬) is the situation where `serde1` is enabled but the user explicitly writes `#[service(derive = [Serialize, Deserialize])]`. That currently causes a conflicting implementation error. I guess I would need to scan through the paths in the meta list and look for `Serialize` or `Deserialize`. Of course, that wouldn't fix someone importing `Serialize` under a different name.

tikue:
>  the situation where serde1 is enabled but the user explicitly writes #[service(derive = [Serialize, Deserialize])]

Oh, this is what I meant to address by my 4th bullet of https://github.com/google/tarpc/issues/436#issuecomment-1961860313: only implicitly derive serialize when `derive = [...]` is not present.

ShaneMurphy2:
> > the situation where serde1 is enabled but the user explicitly writes #[service(derive = [Serialize, Deserialize])]
> 
> Oh, this is what I meant to address by my 4th bullet of [#436 (comment)](https://github.com/google/tarpc/issues/436#issuecomment-1961860313): only implicitly derive serialize when `derive = [...]` is not present.

Gotcha, will work on it.

ShaneMurphy2:
Should be good now, I think I incorrectly updated the CI file though. Either way, do we wanna bikeshed the name a bit? Is `derive` clear? I should also update the documentation for the macro.

ShaneMurphy2:
Cleaned up the implementation a bit.

ShaneMurphy2:
Added some docs and should fix CI 🤞

tests-build/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "tests-build"
```
tikue, line 2: Instead of adding this package, could the tests go in the existing tarpc/tests/compile_fail?

ShaneMurphy2: Ah, assumed `trybuild` tests would be in the plugins crate, my bad. Yes, I believe they can go there.

ShaneMurphy2: Done!

.github/workflows/main.yml
```diff
@@ -56,11 +56,12 @@ jobs:
...
-          cargo test --manifest-path tarpc/Cargo.toml
-          ${{ matrix.serde }} ${{ matrix.tokio }} ${{ matrix.serde-transport }}
-          ${{ matrix.serde-transport-json }} ${{ matrix.serde-transport-bincode }}
-          ${{ matrix.tcp }} ${{ matrix.unix }}
+      - run: |
+          cargo test --manifest-path tarpc/Cargo.toml \
+          ${{ matrix.serde }} ${{ matrix.tokio }} ${{ matrix.serde-transport }} \
+          ${{ matrix.serde-transport-json }} ${{ matrix.serde-transport-bincode }} \
+          ${{ matrix.tcp }} ${{ matrix.unix }} \
+          cargo test --manifest-path tests-build/Cargo.toml $ {{ matrix.serde }}
```
tikue, line 64: this isn't needed now, right?

ShaneMurphy2: Yup, will remove.

tikue:
Still seeing some test failures: 
image

ShaneMurphy2:
> Still seeing some test failures: image

Hm, tests passed locally, and those outputs look the same except for the number of candidate traits. Interesting.

tikue:
> > Still seeing some test failures: image
> 
> Hm, tests passed locally, and those outputs look the same except for the number of candidate traits. Interesting.

sometimes I've seen these changes based on whether the tests were run with nightly vs stable rustc?

ShaneMurphy2:
> > > Still seeing some test failures: image
> > 
> > 
> > Hm, tests passed locally, and those outputs look the same except for the number of candidate traits. Interesting.
> 
> sometimes I've seen these changes based on whether the tests were run with nightly vs stable rustc?

I think it's that the test matrix in CI is enabling features that bring in dependencies that provide a serialize method under a trait. Because of that, the compiler is suggesting more items, causing the output to be different. I could re-generate the stderr files with all feature enabled, but I'm not sure if that would cause CI to fail in a different spot when it toggles the feature off. I think I'll try refactoring the test to use fully qualified syntax to the serialize method so that the error message is not ambiguous.

ShaneMurphy2:
That should fix it, I'm a little worried about how brittle these tests are though. Let me know if you think they're more trouble than they're worth.

tikue:
Oh looks like the CLA hasn't been signed, either? Is that something you can take a look at?

ShaneMurphy2:
> Oh looks like the CLA hasn't been signed, either? Is that something you can take a look at?

Yeah, legal is looking at it. Should just be due diligence and not a problem.

ShaneMurphy2:
> That should fix it, I'm a little worried about how brittle these tests are though. Let me know if you think they're more trouble than they're worth.

Forgot to overwrite the .stderr files 🤦

ShaneMurphy2:
Green!

ShaneMurphy2:
Re: CLA, the legal team is either at a conference or out sick, so it might be a few more days.

ShaneMurphy2:
> Re: CLA, the legal team is either at a conference or out sick, so it might be a few more days.

Approved, just waiting for wheels to turn.

ShaneMurphy2:
Should be good to re-run the pipeline!

plugins/Cargo.toml
```diff
@@ -23,6 +23,7 @@ travis-ci = { repository = "google/tarpc" }
 proc-macro2 = "1.0"
 quote = "1.0"
 syn = { version = "2.0", features = ["full"] }
+proc-macro-warning = "1"
```
tikue, line 26: Just noticed this — are any well-known crates using it? Just want to make sure I do some due diligence before taking on a new (unfamiliar to me) dependency.

ShaneMurphy2: Not off the top of my head, if you know a way to emit a warning without this I'm happy to refactor.

tikue: Looking at the crate, I think the magic is just these lines: https://github.com/ggwpez/proc-macro-warning/blob/970809f551eb78ea003006ef4da0c303ede8501d/proc-macro-warning/src/lib.rs#L260-L272

Basically, it just adds a secret line like this:
```rust
#[deprecated(note = "...")]
const __DEPRECATION_NOTICE: () = ();
let _ = __DEPRECATION_NOTICE;
```

ShaneMurphy2: Ok, want me to remove the dependency and do something like that?

tikue: Yes, I think I'd prefer that, since looks to be just a small amount of code. Thanks very much!

ShaneMurphy2: Will do!

ShaneMurphy2: Done! I like this output better anyways 🙂

tikue (approved):

tikue:
Thanks a lot! This is a great generalization.

PR #448, opened 2024-03-28
Allow futures on Serve and Stub to be Send

ShaneMurphy2:
Not sure if this approach is the right one, so this MR is a draft.

### DIFF
--- a/tarpc/Cargo.toml
+++ b/tarpc/Cargo.toml
@@ -59,6 +59,7 @@ tracing = { version = "0.1", default-features = false, features = [
     "log",
 ] }
 tracing-opentelemetry = { version = "0.22.0", default-features = false }
+trait-variant = "0.1.1"
 opentelemetry = { version = "0.21.0", default-features = false }
 
 
--- a/tarpc/src/client/stub.rs
+++ b/tarpc/src/client/stub.rs
@@ -16,6 +16,7 @@ mod mock;
 /// A connection to a remote service.
 /// Calls the service with requests of type `Req` and receives responses of type `Resp`.
 #[allow(async_fn_in_trait)]
+#[trait_variant::make(TokioStub: Send)]
 pub trait Stub {
     /// The service request type.
     type Req: RequestName;
@@ -40,13 +41,13 @@ where
     }
 }
 
-impl<S> Stub for S
-where
-    S: Serve + Clone,
-{
-    type Req = S::Req;
-    type Resp = S::Resp;
-    async fn call(&self, ctx: context::Context, req: Self::Req) -> Result<Self::Resp, RpcError> {
-        self.clone().serve(ctx, req).await.map_err(RpcError::Server)
-    }
-}
+// impl<S> Stub for S
+// where
+//     S: Serve + Clone,
+// {
+//     type Req = S::Req;
+//     type Resp = S::Resp;
+//     async fn call(&self, ctx: context::Context, req: Self::Req) -> Result<Self::Resp, RpcError> {
+//         self.clone().serve(ctx, req).await.map_err(RpcError::Server)
+//     }
+// }
--- a/tarpc/src/client/stub.rs
+++ b/tarpc/src/client/stub.rs
@@ -2,9 +2,7 @@
 
 use crate::{
     client::{Channel, RpcError},
-    context,
-    server::Serve,
-    RequestName,
+    context, RequestName,
 };
 
 pub mod load_balance;
@@ -40,14 +38,3 @@ where
         Self::call(self, ctx, request).await
     }
 }
-
-// impl<S> Stub for S
-// where
-//     S: Serve + Clone,
-// {
-//     type Req = S::Req;
-//     type Resp = S::Resp;
-//     async fn call(&self, ctx: context::Context, req: Self::Req) -> Result<Self::Resp, RpcError> {
-//         self.clone().serve(ctx, req).await.map_err(RpcError::Server)
-//     }
-// }
--- a/tarpc/src/server.rs
+++ b/tarpc/src/server.rs
@@ -6,6 +6,8 @@
 
 //! Provides a server that concurrently handles many connections sending multiplexed requests.
 
+use crate::client::stub::Stub;
+use crate::client::RpcError;
 use crate::{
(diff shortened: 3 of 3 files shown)


tikue:
Sorry for the delay in reviewing! Will try to look in the next week.

tarpc/src/server.rs
```diff
@@ -66,6 +68,27 @@ impl Config {
     }
 }
 
+/// A [`Stub`] implementation that simply warps a `Serve`.
```
tikue, line 71: Nit: typo

```suggestion
/// A [`Stub`] implementation that simply wraps a `Serve`.
```

tarpc/src/server.rs
```diff
@@ -66,6 +68,27 @@ impl Config {
...
+pub struct ServeStub<S> {
```
tikue, line 72: Maybe keep this type private for now and return an imp Stub? Can always make it public later if needed.

tikue (changes requested):
This looks like a reasonable change to me. Doesn't Serve need to be made a trait variant, too? Also, could you add some tests demonstrating how it will work for users?

ShaneMurphy2:
> This looks like a reasonable change to me.

Perfect, that's what I was looking for by putting up an early draft.
> Doesn't Serve need to be made a trait variant, too?

Probably, was just focusing on my use case a little too much 😅 
> Also, could you add some tests demonstrating how it will work for users?

Will do, as I mentioned this is an early draft to make sure the change makes sense.

tarpc/src/server.rs
ShaneMurphy2: Good idea

ShaneMurphy2:
I'll get back to this in a day or so, thanks for the review!

ShaneMurphy2:
Haven't forgotten, just wrapped up in other stuff. I'll try to get to it soon!

ShaneMurphy2:
Looks like `trait_variant::make` doesn't support default implementations in trait definitions. I could update the macro in that crate, thoughts?

Also, how do you feel about renaming Serve and Stub to LocalServe and LocalStub so that the "default" type is `Send`, seems like the more common use case potentially.

ShaneMurphy2:
I've given it some more thought and I'm wondering if the right move is to instead have a feature flag that annotates all traits with async methods in them with `async_trait`. Thoughts @tikue?

tikue:
I also just saw https://github.com/rust-lang/rfcs/pull/3654. Wonder how long it will take to be available on nightly?

ShaneMurphy2:
> I also just saw [rust-lang/rfcs#3654](https://github.com/rust-lang/rfcs/pull/3654). Wonder how long it will take to be available on nightly?

Looks like even post-implementation the RFC still recommends the pattern provided by `trait_variant`: https://github.com/nikomatsakis/rfcs/blob/rtn/text/0000-return-type-notation.md#rtn-supports-convenient-trait-aliases

0xdeafbeef:
Sorry for ping, any updates?

ShaneMurphy2:
> Sorry for ping, any updates?

I'm currently using my fork's branch `trait-variant` and it seems to work fine. Probably that could be merged soon since it is backwards compatible with the old way, as it only adds `Tokio*` types that bound the futures to be `Send`.

tarpc/src/client/stub.rs
```diff
@@ -16,6 +14,7 @@ mod mock;
 /// A connection to a remote service.
 /// Calls the service with requests of type `Req` and receives responses of type `Resp`.
 #[allow(async_fn_in_trait)]
+#[trait_variant::make(TokioStub: Send)]
```
tikue, line 17: Nit: maybe just call it `SendStub`?

tarpc/Cargo.toml
```diff
@@ -49,7 +49,7 @@ pin-project = "1.0"
 rand = "0.8"
 serde = { optional = true, version = "1.0", features = ["derive"] }
 static_assertions = "1.1.0"
-tarpc-plugins = { path = "../plugins", version = "0.13" }
+tarpc-plugins = { path = "../plugins", version = "0.13", registry = "umbra"}
```
tikue, line 52: Is the `registry = "umbra"` bit supposed to be here?

tarpc/src/server.rs
```diff
@@ -77,6 +100,16 @@ pub trait Serve {
 
     /// Responds to a single request.
     async fn serve(self, ctx: context::Context, req: Self::Req) -> Result<Self::Resp, ServerError>;
+
+    /// Wrap this `Serve` in a type that implements [`Stub`].
+    async fn into_stub(self) -> ServeStub<Self>
```
tikue, line 105: Can you remind me why the blanket impl for Serve impls had to be removed?

tikue (commented):
Hey @ShaneMurphy2 taking another look at this in a while. Are you still interested in getting this merged?

Also, I'm kind of curious about trying out a feature-gated implementation that uses return-type notation rather than `trait_variant`. Some related thoughts on https://github.com/google/tarpc/pull/480

tarpc/src/client/stub.rs
ShaneMurphy2: Sure!

tarpc/Cargo.toml
ShaneMurphy2: Nope, coworker pushed to the wrong branch.

tarpc/src/server.rs
ShaneMurphy2: I think the compiler was complaining about conflicting trait implementations.

ShaneMurphy2:
> Hey @ShaneMurphy2 taking another look at this in a while. Are you still interested in getting this merged?

I'm ambivalent on getting it merged as I think we can do better with the API, and I'm not confident this is the right approach in the general case. It works for our use case and a fork is ok for us for now.

> Also, I'm kind of curious about trying out a feature-gated implementation that uses return-type notation rather than `trait_variant`. Some related thoughts on #480

That could be neat, I agree with your sentiment about being "forward thinking".

In a similar vein with using feature-gating, it could be nice to put the `async_trait` macro on all the traits behind a feature gate. That macro makes async traits very ergonomic for folks who can afford the performance hit.

PR #465, opened 2024-07-30
Fix test stderr and warning in test

flip1995:
This should turn the CI on master green again.

### DIFF
--- a/plugins/tests/service.rs
+++ b/plugins/tests/service.rs
@@ -74,6 +74,7 @@ fn service_with_cfg_rpc() {
 
 #[test]
 fn syntax() {
+    #[tarpc::service]
     trait Syntax {
         #[deny(warnings)]
         #[allow(non_snake_case)]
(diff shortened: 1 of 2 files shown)


plugins/tests/service.rs
```diff
@@ -74,6 +74,7 @@ fn service_with_cfg_rpc() {
 
 #[test]
 fn syntax() {
+    #[tarpc::service]
     trait Syntax {
```
flip1995, line 78: IIUC this test was supposed to test the `service` syntax. But since the attribute was missing it didn't test anything, but just produced a dead_code warning.

tikue: Wouldn't it still be compiled, even if the compiler thinks it's dead code? What happens if a syntax error is added without `#[test]`?

tikue (approved):

tikue:
Thank you!

PR #466, opened 2024-07-30
Fix CI workflow file

flip1995:
The final step didn't include the Clippy and format tests. So even if those failed, PRs could get merged anyway. This now depends on all steps.

Also, the cancel-workflow-action is no longer necessary, as stated in the README of the workflow. GHA has implemented this feature by now.

Also removes the `check` step, as the final test didn't depend on it, it didn't use `-Dwarnings`, and this is also covered by the `clippy` step.

This should now fail CI in the PR, as the clippy check currently fails.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,19 +11,11 @@ on:
 
 name: Continuous Integration
 
-jobs:
-  check:
-    name: Check
-    runs-on: ubuntu-latest
-    steps:
-      - name: Cancel previous
-        uses: styfle/cancel-workflow-action@0.10.0
-        with:
-          access_token: ${{ github.token }}
-      - uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


flip1995:
To test this, the workflow file changes need approval first.

tikue:
Nice fixes, thank you!

Do you want to do also fix the clippy lint in this PR?

flip1995:
Yes, already had the commit ready with the Clippy fix. Just wanted to make sure, that the PR CI now actually fails. 

Pushed the commit. Workflow needs re-approval by you. But this time it should pass.

tikue:
Looks like the MSRV needs to be bumped, as well (which is fine).

flip1995:
Bumped it to 1.65.0, instead of 1.64, what Clippy wanted. 1.65.0 is the MSRV for the latest opentelemetry versions. I'm still planning to update those deps soon :)

PR #467, opened 2024-08-08
Update all tracing dependencies

flip1995:
Those dependencies were really out of date. The opentelemetry-jaeger dependencies was even deprecated. This bumps all those dependencies to the latest version and replaces jaeger with otlp, like recommended by the opentelemetry project.

---

While doing this, I noticed that Clippy didn't run on tests and examples. I changed that in 2f7ef0015d96fd52ccd15e5ce99b2d11a5637af6 and fixed the fallout. Let me know if I should drop this commit or if you want me to open a separate PR for it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -102,7 +102,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
         with:
           components: clippy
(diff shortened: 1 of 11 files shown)


tikue:
Thanks so much!

PR #479, opened 2024-10-09
Drop static lifetime requirement for request name

stevefan1999-personal:
This PR simply remove the 'static requirement so that the registry can designate its request name, in order to revive it.


tarpc/src/lib.rs
```diff
@@ -255,7 +255,6 @@ use std::{any::Any, error::Error, io, sync::Arc, time::Instant};
 /// A message from a client to a server.
 #[derive(Debug)]
 #[cfg_attr(feature = "serde1", derive(serde::Serialize, serde::Deserialize))]
-#[non_exhaustive]
```
tikue, line 258: Was this intentional? Making these types exhaustive commits tarpc to never extending the rpc protocol. I don't think we want to make that commitment.

tikue: Ah I see — yeah for now, can we separate it out to a separate PR? There may be other options we can consider as well, such as a default constructor.

tikue:
Thank you!

PR #480, opened 2024-10-09
Implement an easier way to get Send Serve and Stub

stevefan1999-personal:
This PR improves upon #475 by observing `trait-variant` output and closely imitating the behavior without major breaking changes. There is no need to fill in a lot of unnecessary Send Sync bounds after this which clogged up in the original idea for #475 by using a downgrade trick using extension trait. All existing codes including the examples and the tests compiles without any changes.


tarpc/src/client/stub.rs
```diff
@@ -24,8 +25,28 @@ pub trait Stub {
...
+    fn call(
+        &self,
+        ctx: context::Context,
+        request: Self::Req,
+    ) -> impl Future<Output = Result<Self::Resp, RpcError>>;
+}
+
+/// A connection to a remote service.
+/// Calls the service with requests of type `Req` and receives responses of type `Resp`.
+pub trait SendStub: Send {
```
tikue, line 37: Would it be possible to use return-type notation to avoid duplicating the trait definition? https://github.com/rust-lang/rust/pull/129629

```rs
pub trait SendStub: Stub + Send where <Self as Stub>::call(..): Send {}

impl<S: Stub + Send> SendStub for S where <S as Stub>::call(..): Send {}
```

Benefits of doing it this way:

- The implementations for `Channel` and `Serve` don't need to be duplicated
- the trait method can still be an `async fn`.

I think the same trick can also be applied to `Serve` and the proc macro-generated service trait.

Of course, the downside is this is still an experimental feature. But I think I'd rather think about solving this in a forward-thinking way. We can add a feature gate for it here, as well, so that it's only enabled if compiled with a tarpc Cargo feature that enables unstable features.

tikue (changes requested):

tarpc/src/client/stub.rs
stevefan1999-personal: Will work on it once it is stablized

PR #481, opened 2024-10-09
Remove non_exhaustive for client request and response

stevefan1999-personal:
Split off from #479


tikue:
Thanks for the PR! Thinking about this more, I think I'm okay experimenting with making these exhaustive.

PR #487, opened 2024-10-30
fix test result statement

getong:


tikue:
Thanks! But it looks like you may be on a nightly compiler?

PR #537, opened 2025-10-01
Move INFO level traces to DEBUG

### LINKED ISSUE #536: Make Info-level traces Debug
Hey, before I spend the time to figure out an ergonomic workaround for this locally, would you accept a PR that changes a bunch (maybe all, I haven't looked yet) of the `tracing::info!()` calls to `tracing::debug!()`?

In my experience Info-level traces are usually reserved for the application and library crates stick to Debug (or Trace) for the things you're tracing at Info.

Right now my logs are being flooded by `tarpc::server: ReceiveRequest`, `tarpc::server: BeginRequest`, `tarpc::server: CompleteRequest`, `tarpc::server: BufferResponse`, `tarpc::server: SendResponse`, and their counterparts on the client side.

I'm aware I can just set `RUST_LOG` to `info,tarpc=warn`, but it would be nice to not have to have to adjust tarpc separately from everything else.

Thanks!

azdle:
INFO level traces/logs are usually reserved for the application.

All of the existing INFO level traces were changed, except for those in examples.

resolves #536

### DIFF
--- a/tarpc/src/client.rs
+++ b/tarpc/src/client.rs
@@ -419,7 +419,7 @@ where
         cx: &mut Context<'_>,
     ) -> Poll<Option<Result<DispatchRequest<Req, Resp>, ChannelError<C::Error>>>> {
         if self.in_flight_requests().len() >= self.config.max_in_flight_requests {
-            tracing::info!(
+            tracing::debug!(
                 "At in-flight request capacity ({}/{}).",
                 self.in_flight_requests().len(),
                 self.config.max_in_flight_requests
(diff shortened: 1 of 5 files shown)


azdle:
The clippy error seem to be caused by 73dcdaa, which itself says is fixing clippy errors.

I'm not sure why the `io::Error::other` changes were made, but if I revert those three changes clippy is happy. I've tacked that commit on here, if you want it.

tikue:
I think things should be good once 8655a735eed092264ac87c674626c85e11d8e5f1 is removed

azdle:
Removed, though it looks like main is still failing.

PR #540, opened 2025-11-18
Separate client and server context

axos88:
This is part 2 of the solution to https://github.com/google/tarpc/issues/392, and builds on top of https://github.com/google/tarpc/pull/539.

It separates the client and server contexts, extracts the common part. Only the shared context (curently containing the full old context) is transmitted between the client and server.

This allows the Client and Server contexts to evolve separately, and will be used in a follow-up PR to add extensions that can be used as communication and saving state by the transport, hooks and rpc implementations.

### DIFF
--- a/tarpc/src/client.rs
+++ b/tarpc/src/client.rs
@@ -144,10 +144,7 @@ where
             );
             shared_context.trace_context.new_child()
         });
-        span.record(
-            "rpc.trace_id",
-            tracing::field::display(shared_context.trace_id()),
-        );
+        span.record("rpc.trace_id", tracing::field::display(shared_context.trace_id()), );
         let (response_completion, mut response) = oneshot::channel();
         let request_id =
             u64::try_from(self.next_request_id.fetch_add(1, Ordering::Relaxed)).unwrap();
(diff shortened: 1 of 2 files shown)


axos88:
Yeah, mostly. I am still tryong to figure out how the pieces fit together as well. The semantics is that the client has some stuff that does not relate to the server (think send priority for example), and is kept private to the client, same goes for server,  which may store per request data on how to respond to it (think mqtt response channel).

But now i am also thinking that there might be stuff that may need to be transferred to the server and back, and allow the server to make modifications to it, adding another layer pf comminication besides the actual rpc call.

Think cookies in the http world, we could effectively extend tarpc with sessions, and with that proper authentication, stateful requests, etc.

What do you think? The implementation is not very nice right now, it would be probably better to make the context generic, so that the transport can effectively choose a proper struct instead of an anymap... maybe a trait so that the deadline and trace context can be extracted?... 

Would need to focus uninterrupted on this for a coiple of hours which is not easy, haha.

axos88:
Use this to view the correct diff: https://github.com/axos88/tarpc/compare/make-context-ref-mut...axos88:tarpc:separate-client-and-server-context

PR #542, opened 2025-11-20
Bump dependencies to latest versions

lukecarr:
Bumps dependencies in `Cargo.toml` files, including fixes for breaking changes in upgrading `rand` crate from `0.8` to `0.9`, and bumped actions in GHA workflow files.

Also drops the `patch` component from all dependency versions in `Cargo.toml` files.

This PR closes #529, #530, #533, #534, and #535.

### DIFF
--- a/.github/workflows/pr_review.yml
+++ b/.github/workflows/pr_review.yml
@@ -4,7 +4,7 @@ jobs:
   clippy:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
     - uses: dtolnay/rust-toolchain@stable
       with:
         components: clippy


tarpc/src/trace.rs
```diff
@@ -91,7 +91,7 @@ impl TraceId {
     /// Returns a random trace ID that can be assumed to be globally unique if `rng` generates
     /// actually-random numbers.
     pub fn random<R: Rng>(rng: &mut R) -> Self {
-        TraceId(rng.r#gen::<NonZeroU128>().get())
+        TraceId(rng.random::<NonZeroU128>().get())
```
lukecarr: `r#` qualifier not needed now that the `rand` crate has introduced a `random` function that doesn't mirror the name of a Rust keyword.

PR #554, opened 2026-03-25
Simplify Future for RequestDispatch

stepancheg:
Stop type-erasing `RequestDispatch::terminal_error`.

The field is already on a type generic over the transport, so it can be
`Option<ChannelError<C::Error>>`. That removes `ChannelError::{upcast_any,
downcast}` and the `dyn Any` round-trip in `poll`.

The context is this: `terminal_error` only works for inflight requests, but this
is not enough to provide good diagnostics if we are sending request to a
dispatch which has already failed.

Instead, there can be a field like `Arc<OnceCell<SomeError>>` shared between
dispatch and channel, so that if writing to channel or reading back from oneshot
channel fails, we could attach that error as error source.

So I start by cleaning up existing code.

### DIFF
--- a/tarpc/src/client.rs
+++ b/tarpc/src/client.rs
@@ -19,7 +19,6 @@ use futures::{prelude::*, ready, stream::Fuse, task::*};
 use in_flight_requests::InFlightRequests;
 use pin_project::pin_project;
 use std::{
-    any::Any,
     convert::TryFrom,
     fmt,
     pin::Pin,
@@ -269,7 +268,6 @@ where
             transport: transport.fuse(),
             in_flight_requests: InFlightRequests::default(),
             pending_requests,
-            terminal_error: None,
         },
     }
 }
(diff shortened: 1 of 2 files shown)


stepancheg:
Bump.

stepancheg:
Hello?

tikue:
Sorry for the delay, I'll take a look soon.

stepancheg:
Bump.

tarpc/src/client.rs
```diff
@@ -659,20 +646,13 @@ where
...
+        let result = ready!(self.run(cx));
+        match result {
+            Ok(()) => Poll::Ready(Ok(())),
+            Err(e) => {
                 tracing::debug!("RpcError::Channel");
-                let e: ChannelError<C::Error> = e
-                    .clone()
-                    .downcast()
-                    .expect("Invariant: ChannelError must store a C::Error");
                 ready!(self.shut_down_with_terminal_error(cx, e.clone().upcast_error()));
```
tikue, line 666: I'm not sure this change works, because `shut_down_with_terminal_error` can return `Pending` and then be called again. The request dispatcher closes the pending requests channel and then waits for the other end of the channel to be closed, so `Pending` can be returned if the other end isn't immediately closed.

Sorry that there aren't good tests for this that would have caught this problem earlier. Would you like to try adding a test?

tikue (changes requested):

tikue:
Apologies again for the long delay. Life got in the way.

tarpc/src/client.rs
tikue: > Instead, there can be a field like `Arc<OnceCell<SomeError>>` shared between dispatch and channel, so that if writing to channel or reading back from oneshot channel fails, we could attach that error as error source.

That sounds good too, but won't there be a regression if this PR is merged without the new field?

stepancheg: Good catch, thanks! This PR is now much simpler, only removes `Any` and uses generics. But most of complexity I wanted to remove, stays.

I used AI to add tests that reproduce the problem. Tests are not very readable, not sure how much value in them.

tikue (approved):

tikue:
This PR needs to be rebased onto main, and then it can be merged!

PR #562, opened 2026-06-06
Remove redundant continue in DeriveMeta::parse

enlorik:
The `continue` at the end of the else branch was the last statement in the loop body, so it had no effect.


enlorik:
Opened by mistake, closing.

PR #566, opened 2026-09-17
feat: Make opentelemetry optional

### LINKED ISSUE #564: Make opentelemetry optional
Right now opentelemetry and co are [unconditional dependencies](https://github.com/google/tarpc/blob/main/tarpc/Cargo.toml#L61).

I use tarpc in a project where I just communicate between two services over stdio; I mostly need it for really convenient abstractions and proc macro. Thus, opentelemetry is an overkill for me, and it's basically a dead dependency that I need to worry about.

It would be great if it was put under a feature and made optional.

BTW -- thank you for a great library! I've checked a lot of lightweight RPC implementations in rust and this was the most convenient by far. ❤️

popzxc:
Fixes #564

Not all projects require opentelemetry, and besides being extra code to compile, it also makes supply chain security harder (see [example](https://github.com/AprilNEA/OpenLogi/pull/487)).

I've kept opentelemetry enabled by default so that the change is backward compatible, hope that works.


popzxc:
@tikue any chance you can look into it?

tikue:
Hi, thanks for the contribution! I will try to look at this in the next week or so.

